Issue #1, opened 2020-06-06
New API


lovesegfault:
I find the current API to be a bit weird with the home/system distinction. On top of this the usage of symlinks is somewhat cumbersome. Here's an idea I had:
```nix
{
  persist = {
    directory = "/state";
    environment.directories = [
      "/var/lib/iwd"
      "/var/lib/bluetooth"
    ];
    environment.files = [
      "/etc/machine-id"
    ];
    users.bemeurer.directories = [
      "pictures"
      "src"
    ];
    users.bemeurer.files = [
      ".gist"
      ".zsh_history"
    ];
  };
}
```

For the directories we add a `mkdir` to `system.activationScripts` and a bind-mount to `fileSystems`.

For the files we'd `touch` and bind-mount just the same, which I think @etu noted was possible.

I did a poor's man attempt at this earlier which got lost in a state-mixup, but I had a lot of trouble there with the mountpoints that were supposed to be files ending up as directories somehow.

talyz:
If you don't want to use the `home-manager` module, I guess the only thing missing from the `NixOS` module is handling files outside `/etc`, so that you could do

```nix
{
  environment.persistence."/state" = {
    directories = [
      "/var/lib/iwd"
      "/var/lib/bluetooth"
      "/home/bemeurer/pictures"
      "/home/bemeurer/src"
    ];
    files = [
      "/etc/machine-id"
      "/home/bemeurer/.gist"
      "/home/bemeurer/.zsh_history"
    ];
  };
}
```

We could add `users.<user>` as a shortcut, but it wouldn't be strictly necessary, and possibly a bit confusing when we have the `home-manager` module..

etu:
I think the `users.<user>` part can be quite good to have, because with that we could make sure that the created files are owned by the right user.

lovesegfault:
Yeah, the reason I formatted it like that was to ensure ownership, as opposed to having to do some hacky path-based detection (e.g. assuming all paths starting with `/home/$user` are owned by `$user`)

Another alternative is to get rid of the distinctions altogether and have something like this:
```nix
{ config, ... }: {
  environment.state = {
    directory = "/state";
    entities = [
      { target = "/var/lib/iwd"; type = "d"; }
      { target = "/home/bemeurer/.gist"; owner = config.users.users.bemeurer; type = "f"; }
    ];
  };
}
```

talyz:
[...]

PR #3, opened 2020-06-07
nixos: bind mount with correct permissions

lovesegfault:
This is yet another step towards #1.

With this the full directory structure is replicated with the correct permissions, thus allowing for non-root files and dirs to be bind-mounted. This enforces that the directories and files in the ephemeral storage have exactly matching ownership to their persistent storage correspondents.

I've tested this locally and have managed to bind mount things in my homedir without major issues.

It increases our bash complexity a little bit, but I tried to make the code well documented.

### DIFF
--- a/nixos.nix
+++ b/nixos.nix
@@ -88,15 +88,66 @@ in
     system.activationScripts =
       let
         # Create a directory in persistent storage, so we can bind
-        # mount it.
-        mkDirCreationSnippet = persistentStoragePath: dir:
-          let
-            targetDir = concatPaths [ persistentStoragePath dir ];
-          in
+        # mount it. The directory structure's mode and ownership mirror those of
+        # persistentStoragePath/dir;
+        # TODO: Move this script to it's own file, add CI with shfmt/shellcheck.
+        mkDirWithPerms = persistentStoragePath: dir:
           ''
-            if [[ ! -e "${targetDir}" ]]; then
-                mkdir -p "${targetDir}"
+            # Given a source directory, /source, and a target directory,
+            # /target/foo/bar/bazz, we want to "clone" the target structure
+            # from source into the target. Essentially, we want both
+            # /source/target/foo/bar/bazz and /target/foo/bar/bazz to exist
+            # on the filesystem. More concretely, we'd like to map
+            # /state/etc/ssh/example.key to /etc/ssh/example.key
+            #
+            # To achieve this, we split the target's path into parts -- target, foo,
+            # bar, bazz -- and iterate over them while accumulating the path
+            # (/target/, /target/foo/, /target/foo/bar, and so on); then, for each of
+            # these increasingly qualified paths we:
+            #
+            #   1. Ensure both /source/qualifiedPath and qualifiedPath exist
+            #   2. Copy the ownership of the source path into the target path
+            #   3. Copy the mode of the source path into the target path
+            (
+            # capture the nix vars into bash to avoid escape hell
+            sourceBase="${persistentStoragePath}"
+            target="${dir}"
+
+            # trim trailing slashes the root of all evil
+            sourceBase="''${sourceBase%/}"
+            target="''${target%/}"
+
+            # check that the source exists, if it doesn't exit the underlying
+            # scope (the activation script will continue)
+            realSource="$(realpath "$sourceBase$target")"
+            if [ ! -d "$realSource" ]; then
+                printf "\e[1;31mBind source '%s' does not exist; it will be created for you.\e[0m\n" "$realSource"
             fi
+
+            # iterate over each part of the target path, e.g. var, lib, iwd
+            previousPath="/"
+            for pathPart in $(echo "$target" | tr "/" " "); do
+              # construct the incremental path, e.g. /var, /var/lib, /var/lib/iwd
+              currentTargetPath="$previousPath$pathPart/"
+
+              # construct the source path, e.g. /state/var, /state/var/lib, ...
+              currentSourcePath="$sourceBase$currentTargetPath"
+
+              # create the source and target directories if they don't exist
+              [ -d "$currentSourcePath" ] || mkdir "$currentSourcePath"
+              [ -d "$currentTargetPath" ] || mkdir "$currentTargetPath"
+
+              # resolve the source path to avoid symlinks
+              currentRealSourcePath="$(realpath "$currentSourcePath")"
+
+              # synchronize perms between source and target
+              chown --reference="$currentRealSourcePath" "$currentTargetPath"
+              chmod --reference="$currentRealSourcePath" "$currentTargetPath"
+
+              # lastly we update the previousPath to continue down the tree
+              previousPath="$currentTargetPath"
+            done
+            )
           '';
 
         # Build an activation script which creates all persistent
@@ -105,7 +156,7 @@ in
           nameValuePair
             "createDirsIn-${replaceStrings [ "/" "." " " ] [ "-" "" "" ] persistentStoragePath}"
             (noDepEntry (concatMapStrings
-              (mkDirCreationSnippet persistentStoragePath)
+              (mkDirWithPerms persistentStoragePath)
               cfg.${persistentStoragePath}.directories
             ));
       in


etu:
I've encountered some really bad issues with this when trying it out.

I'm not sure the model with looking at the persistent stored file works fully if you happen to point at a symlink...

So this would make more sense to do more if we had a `users` api that took the username into account.

And I think the points that @talyz bring up here is important and good: https://github.com/nix-community/impermanence/issues/1#issuecomment-640484933

So I think this was a fun experiment that brought us an assertion about `neededForBoot`, which is great and well on the way to be implemented in #5

lovesegfault:
I think it works if we add a call to `realpath` to resolve the reference given to chmod/chown. Let me give it a shot.

lovesegfault:
@etu try it out :)

I'd like to get this merged into `next` and then we can remove the bind mount for files in a subsequent PR.

lovesegfault:
@talyz @etu Please take a new look, I've rebased this onto master

nixos.nix
```diff
@@ -90,13 +90,67 @@ in
...
+            # /state/etc/ssh/example.key to /etc/ssh/example.key
+            #
+            # To do this we split the target's path into parts (target, foo,
+            # bar, bazz) and iterate over them while accumulating the path
+            # (/target/, /target/foo/, /target/foo/bar, ...) and for each of
+            # these incresingly qualified paths we:
+            #
+            #  1. Enforce /source/qualifiedPath and qualifiedPath exists
+            #  2. Copy the ownership of the source path into the target path
+            #  3. Copy the mode of the source path into the target path
```
cole-h, line 108: "we want to get both /source/target/foo/bar/bazz [1] and /target/foo/bar/bazz to exist" -- shouldn't [1] already exist?

---

```suggestion
            # Given a source directory, /source, and a target directory,
            # (/target/foo/bar/bazz, we want to "clone" the target structure
            # from source into the target. Essentially, we want to get both
            # /source/target/foo/bar/bazz and /target/foo/bar/bazz to exist
            # in the filesystem. More concretely, we'd like to map
            # /state/etc/ssh/example.key to /etc/ssh/example.key
            #
            # To achieve this, we split the target's path into parts -- target, foo,
            # bar, bazz -- and iterate over them while accumulating the path
            # (/target/, /target/foo/, /target/foo/bar, and so on); then, for each of
            # these increasingly qualified paths we:
            #
            #   1. Ensure both /source/qualifiedPath and qualifiedPath exist
            #   2. Copy the ownership of the source path into the target path
            #   3. Copy the mode of the source path into the target path
```

nixos.nix
```diff
@@ -90,13 +90,67 @@ in
...
         mkDirCreationSnippet = persistentStoragePath: dir:
-          let
-            targetDir = concatPaths [ persistentStoragePath dir ];
-          in
           ''
-            if [[ ! -e "${targetDir}" ]]; then
-                mkdir -p "${targetDir}"
-            fi
+            # Given a source directory, /source, and a target directory,
+            # (/target/foo/bar/bazz, we want to "clone" the target structure
```
cole-h, line 95:
```suggestion
            # /target/foo/bar/bazz, we want to "clone" the target structure
```

Oops, I missed this paren.

nixos.nix
```diff
@@ -90,13 +90,67 @@ in
...
+            # from source into the target. Essentially, we want to get both
+            # /source/target/foo/bar/bazz and /target/foo/bar/bazz to exist
+            # in the filesystem. More concretely, we'd like to map
```
cole-h, line 98: I think you missed this question from my last suggestion (easy to miss with that giant wall of text suggestion):

"we want to get both /source/target/foo/bar/bazz [1] and /target/foo/bar/bazz [2] to exist" -- shouldn't [1] already exist, if we're wanting to map [1] to [2]?

---

While I'm here:

```suggestion
            # from source into the target. Essentially, we want both
            # /source/target/foo/bar/bazz and /target/foo/bar/bazz to exist
            # on the filesystem. More concretely, we'd like to map
```

lovesegfault:
Ping @talyz @etu

talyz:
Since you're not really doing any advanced splicing of nix strings into the bash script, I would suggest making the script take the only two values spliced as arguments instead and either create the script inline or move it out of the file to avoid bloating the activation script too much.

nixos.nix
```diff
@@ -90,13 +90,67 @@ in
         # Create a directory in persistent storage, so we can bind
         # mount it.
         mkDirCreationSnippet = persistentStoragePath: dir:
```
talyz, line 92: The name and the comment should to be updated to reflect the new behavior.

nixos.nix
```diff
@@ -90,13 +90,67 @@ in
...
+            previousPath="/"
+            for pathPart in $(echo "$target" | tr "/" " "); do
+              # construct the incremental path, e.g. /var, /var/lib, /var/lib/iwd
+              currentTargetPath="$previousPath$pathPart/"
+
+              # construct the source path, e.g. /state/var, /state/var/lib, ...
+              currentSourcePath="$sourceBase$currentTargetPath"
+
+              if [ ! -d "$currentSourcePath" ]; then
+                printf "Bind source '%s' does not exist, creating it\n" "$currentSourcePath"
```
talyz, line 128: Won't this print on every part of the path? That feels a little too verbose. It also only does this for the source, not the target..

lovesegfault:
> Since you're not really doing any advanced splicing of nix strings into the bash script, I would suggest making the script take the only two values spliced as arguments instead and either create the script inline or move it out of the file to avoid bloating the activation script too much.

I plan to move the script out of the file and add a shellcheck step to CI to make sure it's valid. I just want to do it in a subsequent PR.

nixos.nix
lovesegfault: Done

nixos.nix
lovesegfault: I've refactored this behavior, please take a new look

lovesegfault:
> I'm still a bit unclear on the use case and purpose of this, though. Have I understood it correctly that this is to account for when the source path has stricter permissions set somewhere prior to the bound directory?

Without this:
Suppose you have `/state/foo/bar/` where `foo/` and `bar/` are owned by user alice. You then choose to bind-mount `/foo/bar/`, but the code does a simple `mkdir -p` as root resulting in `/foo/` owned by `root`, and `/bar/` originally owned by `root`, but then by `alice` when the bind mount happens. Regardless, `alice` can't access `/foo/bar/` now because `/foo/` is owned by _not her_.

With this:
Suppose you have `/state/foo/bar/` where `foo/` and `bar/` are owned by user alice. You then choose to bind mount `/foo/bar/`, the code correctly transverses that path and creates an identical structure, which identical modes and ownership. `alice` now owns `/foo/` and `/foo/bar/` and her bind mount works as expected.

nixos.nix
```diff
@@ -117,22 +120,26 @@ in
...
             target="''${target%/}"
 
+            (
+            # check that the source exists
+            realSource="$(realpath "$sourceBase$target")"
+            if [ ! -d "$realSource" ]; then
+                printf "\e[1;31mBind source '%s' does not exist!\e[0m\n" "$realSource"
+            fi
+            )
+            (
```
cole-h: Would improve readability to indent all the stuff inside each scope to match its nesting level.

Also:
```suggestion
            )
            
            (
```

talyz:
> Without this:
> Suppose you have `/state/foo/bar/` where `foo/` and `bar/` are owned by user alice. You then choose to bind mount `/foo/bar/`, but the code does a simple `mkdir -p` as root resulting in `/foo/` owned by `root`, and `/bar/` originally owned by `root`, but then by `alice` when the bind mount happens. Regardless, `alice` can't access `/foo/bar/` now because `/foo/` is owned by _not her_.

Well, she can access it and edit files in it, since the default permissions allow it, even if `/foo` happens to be owned by `root`. I have this setup for my `/etc/nixos` directory, and it works fine. She wouldn't be able to put files in `/foo`, though, which this should fix.

nixos.nix
```diff
@@ -88,15 +88,66 @@ in
...
+            # trim trailing slashes the root of all evil
+            sourceBase="''${sourceBase%/}"
+            target="''${target%/}"
+
+            # check that the source exists, if it doesn't exit the underlying
+            # scope (the activation script will continue)
+            realSource="$(realpath "$sourceBase$target")"
+            if [ ! -d "$realSource" ]; then
+                printf "\e[1;31mBind source '%s' does not exist!\e[0m\n" "$realSource"
+                exit 1
```
talyz, line 125: Wait, why fail instead of just creating the directory? :confused:

lovesegfault: True, I've reverted this. It now warns if the source doesn't exist, but continues to create it.

nixos.nix
```diff
@@ -121,8 +121,7 @@ in
             # scope (the activation script will continue)
             realSource="$(realpath "$sourceBase$target")"
             if [ ! -d "$realSource" ]; then
-                printf "\e[1;31mBind source '%s' does not exist!\e[0m\n" "$realSource"
-                exit 1
+                printf "\e[1;31mBind source '%s' does not exist! We'll create it.\e[0m\n" "$realSource"
```
cole-h, line 124:
```suggestion
                printf "\e[1;31mBind source '%s' does not exist; it will be created for you.\e[0m\n" "$realSource"
```

nixos.nix
```diff
@@ -88,15 +88,66 @@ in
...
+            for pathPart in $(echo "$target" | tr "/" " "); do
+              # construct the incremental path, e.g. /var, /var/lib, /var/lib/iwd
+              currentTargetPath="$previousPath$pathPart/"
+
+              # construct the source path, e.g. /state/var, /state/var/lib, ...
+              currentSourcePath="$sourceBase$currentTargetPath"
+
+              # create the source and target directories
+              mkdir "$currentSourcePath" || true
+              mkdir "$currentTargetPath" || true
```
cole-h, line 138: Should these be `-p`, e.g. when some or all of the parents of `{/state,}/var/lib/iwd` don't exist?

lovesegfault: It shouldn't have to be `-p` because we're traversing both the source and target paths. In other words, for `->` meaning "apply permissions and ownership clone" we do;
1. `/state/var -> /var`
2. `/state/var/lib -> /var/lib`
3. `/state/var/lib/iw -> /var/lib/iwd`

So there's no need for `-p` because we manually go through the parents.

cole-h: Ah that's right, I forgot we were looping over the components here... lol. Thanks for correcting me.

talyz:
I just tried this out and it seems to work; I don't have anything in my setup that actually tests the permission reassignment, but it doesn't break anything for me. I'm getting some error messages on boot, though:
```
Jun 14 22:11:32 flora stage-2-init: mkdir: cannot create directory '/persistent/var/': File exists
Jun 14 22:11:32 flora stage-2-init: mkdir: cannot create directory '/var/': File exists
Jun 14 22:11:32 flora stage-2-init: mkdir: cannot create directory '/persistent/var/log/': File exists
Jun 14 22:11:32 flora stage-2-init: mkdir: cannot create directory '/var/log/': File exists
Jun 14 22:11:32 flora stage-2-init: mkdir: cannot create directory '/persistent/var/': File exists
Jun 14 22:11:32 flora stage-2-init: mkdir: cannot create directory '/var/': File exists
Jun 14 22:11:32 flora stage-2-init: mkdir: cannot create directory '/persistent/var/lib/': File exists
Jun 14 22:11:32 flora stage-2-init: mkdir: cannot create directory '/var/lib/': File exists
Jun 14 22:11:32 flora stage-2-init: mkdir: cannot create directory '/persistent/var/lib/bluetooth/': File exists
Jun 14 22:11:32 flora stage-2-init: mkdir: cannot create directory '/persistent/var/': File exists
Jun 14 22:11:32 flora stage-2-init: mkdir: cannot create directory '/var/': File exists
Jun 14 22:11:32 flora stage-2-init: mkdir: cannot create directory '/persistent/var/lib/': File exists
Jun 14 22:11:32 flora stage-2-init: mkdir: cannot create directory '/var/lib/': File exists
Jun 14 22:11:32 flora stage-2-init: mkdir: cannot create directory '/persistent/var/lib/libvirt/': File exists
Jun 14 22:11:32 flora stage-2-init: mkdir: cannot create directory '/persistent/var/': File exists
Jun 14 22:11:32 flora stage-2-init: mkdir: cannot create directory '/var/': File exists
Jun 14 22:11:32 flora stage-2-init: mkdir: cannot create directory '/persistent/var/lib/': File exists
Jun 14 22:11:32 flora stage-2-init: mkdir: cannot create directory '/var/lib/': File exists
Jun 14 22:11:32 flora stage-2-init: mkdir: cannot create directory '/persistent/var/lib/docker/': File exists
Jun 14 22:11:32 flora stage-2-init: mkdir: cannot create directory '/persistent/var/': File exists
Jun 14 22:11:32 flora stage-2-init: mkdir: cannot create directory '/var/': File exists
Jun 14 22:11:32 flora stage-2-init: mkdir: cannot create directory '/persistent/var/lib/': File exists
Jun 14 22:11:32 flora stage-2-init: mkdir: cannot create directory '/var/lib/': File exists
Jun 14 22:11:32 flora stage-2-init: mkdir: cannot create directory '/persistent/var/lib/systemd/': File exists
Jun 14 22:11:32 flora stage-2-init: mkdir: cannot create directory '/persistent/var/lib/systemd/coredump/': File exists
Jun 14 22:11:32 flora stage-2-init: mkdir: cannot create directory '/persistent/etc/': File exists
Jun 14 22:11:32 flora stage-2-init: mkdir: cannot create directory '/etc/': File exists
Jun 14 22:11:32 flora stage-2-init: mkdir: cannot create directory '/persistent/etc/NetworkManager/': File exists
Jun 14 22:11:32 flora stage-2-init: mkdir: cannot create directory '/persistent/etc/NetworkManager/system-connections/': File exists
```
Could you add a check for the existence of the directory before running `mkdir`?

lovesegfault:
@talyz Done!

nixos.nix
```diff
@@ -88,15 +88,66 @@ in
...
         # Create a directory in persistent storage, so we can bind
-        # mount it.
-        mkDirCreationSnippet = persistentStoragePath: dir:
-          let
-            targetDir = concatPaths [ persistentStoragePath dir ];
-          in
+        # mount it. The directory structure's mode and ownership mirror those of
+        # persistentStoragePath/dir;
+        # TODO: Move this script to it's own file, add CI with shfmt/shellcheck.
+        mkDirWithModeAndPerms = persistentStoragePath: dir:
```
cole-h, line 94:
```suggestion
        mkDirWithPerms = persistentStoragePath: dir:
```

I'd argue that having both "mode" and "perms" is redundant, since the mode is (IMO) a way to set permissions ("Can others read, write, or execute this? Yes; I permit others to read and execute this."). The function name becomes less of a mouthful if we shorten it to the above suggestion.

lovesegfault: Done in separate commit.

talyz:
@lovesegfault Looks good! Great work! Before we merge this, can you please squash this down to one commit? Looking at the changes aggregated, it's pretty clear what is done - it's one connected thing and you provide good comments in the code, so I don't think it needs to be more than one commit in its final form.

lovesegfault:
@talyz Done!

PR #5, opened 2020-06-07
nixos: Add comments and neededForBoot assert

talyz:
Add some comments and the `neededForBoot` assertion from @etu's `impermanence.nix` work.

@etu: If you want to, you can change the author of the assertion commit, since you did the work for it. :)

### DIFF
--- a/nixos.nix
+++ b/nixos.nix
@@ -104,6 +104,7 @@ in
     assertions =
       let
         files = concatMap (p: p.files or [ ]) (attrValues cfg);
+        markedNeededForBoot = cond: fs: (config.fileSystems.${fs}.neededForBoot == cond);
       in
       [
         {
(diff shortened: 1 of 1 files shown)


nixos.nix
```diff
@@ -115,6 +128,24 @@ in
             ''
               environment.persistence.files:
                   Currently, only files in /etc are supported.
+
+                  Please fix / remove the following paths:
```
cole-h, line 132:
```suggestion
                  Please fix or remove the following paths:
```

I feel like using `/` for "or" here is needlessly confusing (e.g. it may be mistaken for the root of a filesystem).

nixos.nix
```diff
@@ -115,6 +128,24 @@ in
...
+          message =
+            let
+              offenders = filter (markedNeededForBoot false) persistentStoragePaths;
+            in
+            ''
+              environment.persistence:
+                  All filesystems used for persistent storage must
+                  have the flag neededForBoot set to true.
+
                   Please fix / remove the following paths:
```
cole-h, line 149:
```suggestion
                  Please fix or remove the following paths:
```

Same here.

nixos.nix
talyz: Ah, very true.

PR #6, opened 2020-06-08
LICENSE: init

### LINKED ISSUE #4: Set a license
The repo currently doesn't have a license, but we should set one.

There is currently code from myself, @talyz, and @etu. Do we all agree on licensing our contributions under BSD-3-Clause?

c.f. https://opensource.org/licenses/BSD-3-Clause

lovesegfault:
Closes #4.

Going forward with BSD-3-Clause since no one complained.

@etu @talyz Please review to make sure you're okay with licensing your work under BSD-3.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2020 Nix Community Projects
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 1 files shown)


Mic92:
BSD3 seems fine. I choose MIT for [NUR](https://github.com/nix-community/NUR) because it is the same license we use for nixpkgs.

grahamc:
Good point. If we wanted to upstream this to NixOS, it would need to be MIT.

talyz:
Maybe we should go with MIT too, then? :)

etu:
MIT is fine as well :)

lovesegfault:
Merging as MIT :)

PR #8, opened 2020-06-12
nixos: allow files to be symlinked anywhere

lovesegfault:
This removes the current restriction on `files` needing to exist in `/etc` and allows them to be symlinked anywhere on disk.

### DIFF
--- a/nixos.nix
+++ b/nixos.nix
@@ -39,30 +39,6 @@ in
   };
 
   config = {
-    environment.etc =
-      let
-        link = file:
-          pkgs.runCommand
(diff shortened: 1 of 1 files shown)


lovesegfault:
@talyz review ping :)

talyz:
@lovesegfault Sorry I haven't reviewed this earlier; I believe we discussed this approach in another issue, so I thought I already had.

My issue with this is that it doesn't clean up symlinks when you switch to a new generation. This is something that we get for free with `environment.etc`, but we should also be able to use systemd units, as we do currently with the the home-manager bind mounts. That would also let us dynamically choose whether we symlink or bind mount files (symlink if the real file doesn't exist, bind mount if it does).

PR #16, opened 2020-08-17
Make impermanence a flake

danieldk:
Thanks for this very nice module! This small change makes this repository usable as a flake.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,7 @@
+{
+  inputs.nixpkgs.url = "github:nixos/nixpkgs/nixos-20.09-small";
+
+  outputs = { self, nixpkgs }: {
+    nixosModules.impermanence = import ./nixos.nix;
+  };
+}
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,27 @@
+{
+  "nodes": {
+    "nixpkgs": {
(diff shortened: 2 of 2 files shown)


flake.nix
```diff
@@ -0,0 +1,8 @@
+{
+  inputs.nixpkgs.url = "nixpkgs/nixos-20.03-small";
+
+  outputs = { self, nixpkgs }: {
+    nixosModule = self.nixosModules.impermanence;
+    nixosModules.impermanence = import ./nixos.nix;
```
talyz, line 6: Are both necessary? The wiki entry makes it sound like `nixosModules` is only needed if multiple modules are provided.

talyz (commented):
Thanks a lot for this! I have yet to try out flakes, so I obviously haven't tried this, but it looks good to me: I only have one question.

bbigras:
Any progress on this?

cole-h:
I took the liberty of resolving the above question by using `nixosModules.impermanence` over `nixosModule` (this is what e.g. `home-manager` and `Nixpkgs` use).

I also updated the `nixpkgs` input to point to 20.09, since that is the new stable. PTAL.

bbigras:
Thanks everyone!

PR #19, opened 2020-09-19
home-manager: Make quoting and grepping more robust

talyz:
Try to more properly escape paths using `escapeShellArg` and use `grep -F` to avoid accidental regexes.

### DIFF
--- a/home-manager.nix
+++ b/home-manager.nix
@@ -78,9 +78,8 @@ in
             name = "bindMount-${sanitizeName targetDir}";
             startScript = pkgs.writeShellScript name ''
               set -eu
-              if ! ${pkgs.utillinux}/bin/mount | ${pkgs.gnugrep}/bin/grep -F ${mountPoint}' ' \
-                 && ! ${pkgs.utillinux}/bin/mount | ${pkgs.gnugrep}/bin/grep -F ${mountPoint}/; then
-                  ${pkgs.bindfs}/bin/bindfs -f --no-allow-other ${targetDir} ${mountPoint}
+              if ! mount | grep -F ${mountPoint}' ' && ! mount | grep -F ${mountPoint}/; then
+                  bindfs -f --no-allow-other ${targetDir} ${mountPoint}
               else
                   echo "There is already an active mount at or below ${mountPoint}!" >&2
                   exit 1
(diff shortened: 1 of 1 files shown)


home-manager.nix
```diff
@@ -73,28 +73,29 @@ in
...
+            targetDir = escapeShellArg (concatPaths [ persistentStoragePath dir ]);
+            mountPoint = escapeShellArg (concatPaths [ config.home.homeDirectory mountDir ]);
             name = "bindMount-${sanitizeName targetDir}";
             startScript = pkgs.writeShellScript name ''
               set -eu
-              if ! ${pkgs.utillinux}/bin/mount | ${pkgs.gnugrep}/bin/grep -E '${mountPoint}( |/)'; then
-                  ${pkgs.bindfs}/bin/bindfs -f --no-allow-other "${targetDir}" "${mountPoint}"
+              if ! ${pkgs.utillinux}/bin/mount | ${pkgs.gnugrep}/bin/grep -F ${mountPoint}' ' \
+                 && ! ${pkgs.utillinux}/bin/mount | ${pkgs.gnugrep}/bin/grep -F ${mountPoint}/; then
+                  ${pkgs.bindfs}/bin/bindfs -f --no-allow-other ${targetDir} ${mountPoint}
```
lovesegfault, line 83: I feel like it'd be better to bind these binaries to variables in Nix, like `mount = "${pkgs.utillinux}/bin/mount";` and then using them as `${mount}` instead of having the long path everywhere.

home-manager.nix
```diff
@@ -73,28 +73,29 @@ in
...
               else
-                  echo "There is already an active mount at or below ${mountPoint}!" >&2
+                  echo "There is already an active mount at or below "${mountPoint}"!" >&2
```
cole-h, line 85: Maybe it's just me, but I think rendering the single quotes is desirable (e.g. drop the added double quotes). Take, for example: a path with a space in it, `/root/some/mount point`. With these double quotes, the message will display as:

```
There is already an active mount at or below /root/some/mount point!
```

But without these double quotes, the message will display as:

```
There is already an active mount at or below '/root/some/mount point'!
```

home-manager.nix
talyz: Yeah, that sounds a lot cleaner. :)

home-manager.nix
talyz: Yep, that looks better; I'll drop the added double quotes.

cole-h (approved):
Diff LGTM (but I can't attest to functionality).

lovesegfault (approved):
LGTM

Issue #26, opened 2020-11-28
Unused flake input


linyinfeng:
Is this flake input necessary?

https://github.com/nix-community/impermanence/blob/a7169e575a845de6b3b8280a708da7e2d7403950/flake.nix#L2

cole-h:
Maybe not. Fixed in #27, PTAL.

PR #27, opened 2020-11-29
flake: remove unused nixpkgs input

cole-h:
Packages are retrieved from the `pkgs` module argument, rather than
Nixpkgs itself.

---

Fixes https://github.com/nix-community/impermanence/issues/26.

`nix flake check` passes, but I'd appreciate testing to verify things still work properly.

cc @linyinfeng


linyinfeng:
@cole-h I have tested it on my laptop with `nix flake update --override-input impermanence github:cole-h/impermanence/remove-unused-input` and `nixos-rebuild`, and it works well.

nix does not generate `flake.lock` if there are no inputs. Maybe there is no need to add `flake.lock` to `.gitignore`?

cole-h:
Indeed, I didn't notice. Fixed, thanks.

PR #29, opened 2020-12-13
Create directories file

talyz:
Extract the `createDirectories` script out into its own file and add CI checks. Also, run the script for directories of files in persistent storage, not just pure directories.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,21 +1,41 @@
 name: CI
 on: [push, pull_request]
 jobs:
-  parsing:
+  nix_parsing:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
       - uses: cachix/install-nix-action@v12
-      - name: Check Nix Parsing
+      - name: Check Nix parsing
         run: |
           find . -name "*.nix" -exec nix-instantiate --parse --quiet {} >/dev/null +
-  format:
+  nix_formatting:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
       - uses: cachix/install-nix-action@v12
         with:
           nix_path: nixpkgs=channel:nixos-unstable
-      - name: Check Nix Formatting
+      - name: Check Nix formatting
         run: |
           nix run nixpkgs.nixpkgs-fmt -c nixpkgs-fmt --check .
+  shell_formatting:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: cachix/install-nix-action@v12
+        with:
+          nix_path: nixpkgs=channel:nixos-unstable
+      - name: Check shell script formatting
+        run: |
+          find . -name "*.*sh" -exec nix run nixpkgs.shfmt -c shfmt -i 4 -d {} +
+  shell_error_checking:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: cachix/install-nix-action@v12
+        with:
+          nix_path: nixpkgs=channel:nixos-unstable
+      - name: Check for shell script errors
+        run: |
+          find . -name "*.*sh" -exec nix run nixpkgs.shellcheck -c shellcheck {} +
--- a/create-directories.bash
+++ b/create-directories.bash
@@ -1,0 +1,65 @@
+#!/usr/bin/env bash
+
+set -o nounset
+set -o errexit
+set -o pipefail
+set -o errtrace
+
+trap 'echo Error when executing ${BASH_COMMAND} at line ${LINENO}! >&2' ERR
+
+# Given a source directory, /source, and a target directory,
+# /target/foo/bar/bazz, we want to "clone" the target structure
+# from source into the target. Essentially, we want both
+# /source/target/foo/bar/bazz and /target/foo/bar/bazz to exist
+# on the filesystem. More concretely, we'd like to map
+# /state/etc/ssh/example.key to /etc/ssh/example.key
+#
+# To achieve this, we split the target's path into parts -- target, foo,
+# bar, bazz -- and iterate over them while accumulating the path
+# (/target/, /target/foo/, /target/foo/bar, and so on); then, for each of
+# these increasingly qualified paths we:
+#   1. Ensure both /source/qualifiedPath and qualifiedPath exist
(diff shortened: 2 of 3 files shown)


create-directories.bash
```diff
@@ -0,0 +1,56 @@
...
+# from source into the target. Essentially, we want both
+# /source/target/foo/bar/bazz and /target/foo/bar/bazz to exist
+# on the filesystem. More concretely, we'd like to map
+# /state/etc/ssh/example.key to /etc/ssh/example.key
+#
+# To achieve this, we split the target's path into parts -- target, foo,
+# bar, bazz -- and iterate over them while accumulating the path
+# (/target/, /target/foo/, /target/foo/bar, and so on); then, for each of
+# these increasingly qualified paths we:
+#
```
lovesegfault, line 14:
```suggestion
```

create-directories.bash
```diff
@@ -0,0 +1,56 @@
...
+#   1. Ensure both /source/qualifiedPath and qualifiedPath exist
+#   2. Copy the ownership of the source path into the target path
```
lovesegfault, line 16:
```suggestion
#   2. Copy the ownership of the source path to the target path
```

create-directories.bash
```diff
@@ -0,0 +1,56 @@
...
+#   3. Copy the mode of the source path into the target path
```
lovesegfault, line 17:
```suggestion
#   3. Copy the mode of the source path to the target path
```

create-directories.bash
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env bash
+
```
lovesegfault, line 2:
```suggestion
set -o errexit
set -o nounset
set -o pipefail
```

create-directories.bash
```diff
@@ -0,0 +1,56 @@
...
+# To achieve this, we split the target's path into parts -- target, foo,
+# bar, bazz -- and iterate over them while accumulating the path
+# (/target/, /target/foo/, /target/foo/bar, and so on); then, for each of
+# these increasingly qualified paths we:
+#
+#   1. Ensure both /source/qualifiedPath and qualifiedPath exist
+#   2. Copy the ownership of the source path into the target path
+#   3. Copy the mode of the source path into the target path
+
+# Get inputs from command line arguments
```
lovesegfault, line 19:
```suggestion
# Get inputs from command line arguments
if [ "$#" -neq 2 ]; then
    printf "\e[1;31mError: 'create-directories.bash' requires *two* args.\e[0m\n"
    exit 1
fi
```

create-directories.bash
```diff
@@ -0,0 +1,56 @@
...
+sourceBase="$1"
+target="$2"
+
+# trim trailing slashes the root of all evil
+sourceBase="${sourceBase%/}"
+target="${target%/}"
+
+# check that the source exists, if it doesn't exit the underlying
+# scope (the activation script will continue)
```
lovesegfault, line 28: This comment seems out-of-date?

nixos.nix
```diff
@@ -156,14 +102,18 @@ in
...
+          in
+            nameValuePair
+              "createDirsIn-${replaceStrings [ "/" "." " " ] [ "-" "" "" ] persistentStoragePath}"
+              (noDepEntry (concatMapStrings
+                (mkDirWithPerms persistentStoragePath)
+                (directories ++ files)
+              ));
       in
-      listToAttrs (map mkDirCreationScriptForPath persistentStoragePaths);
+        listToAttrs (map mkDirCreationScriptForPath persistentStoragePaths);
```
lovesegfault, line 116: Nice, this is easier to read

lovesegfault (changes requested):
Awesome :)

.github/workflows/ci.yml
```diff
@@ -6,16 +6,36 @@ jobs:
...
+  # format:
+  #   runs-on: ubuntu-latest
+  #   steps:
+  #     - uses: actions/checkout@v2
+  #     - uses: cachix/install-nix-action@v12
+  #       with:
+  #         nix_path: nixpkgs=channel:nixos-unstable
+  #     - name: Check Nix formatting
+  #       run: |
+  #         nix run nixpkgs.nixpkgs-fmt -c nixpkgs-fmt --check .
```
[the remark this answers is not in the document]
lovesegfault: Oh, just saw the PR description. Let me see if I can figure out what it's doing wrong/weird.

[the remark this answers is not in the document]
lovesegfault: Seems like it's just enforcing no indent following an `in` keyword? Like:
```
...
in
  foo
```
it wants to be 
```
...
in
foo
```

While I, personally, prefer the indent following `in` I'd still rather have some authoritative formatting check that I disagree with on occasion than no formatting check.

Not willing to die on this hill though :D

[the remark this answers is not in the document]
talyz: Ah, never mind, I used `nixfmt` by mistake. It wanted to reformat the whole file, which seemed a bit excessive ;)

create-directories.bash
talyz: Added this and a `trap` to provide a better error message.

create-directories.bash
talyz: Added this, but with a `[[` test instead, which is easier to read and harder to mess up. I changed all other tests from `[` to `[[` too, to standardize. Also, instead of making the text red, I print it to `stderr`.

create-directories.bash
talyz: Yeah, it should be up-to-date now. I also added a note for the user to check permissions on the auto-created directories and made it print to `stderr` - while not strictly an error, this makes it easier to find in the log.

create-directories.bash
lovesegfault: Works for me :)

.github/workflows/ci.yml
```diff
@@ -6,7 +6,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: cachix/install-nix-action@v12
-      - name: Check Nix Parsing
+      - name: Check Nix parsing
         run: |
           find . -name "*.nix" -exec nix-instantiate --parse --quiet {} >/dev/null +
   format:
```
lovesegfault, line 12: Maybe call it `nix_formatting` now that there's a `shell_formatting`?

lovesegfault (approved):
LGTM!

PR #38, opened 2021-01-23
Fix bindfs unmounts and remounts

### LINKED ISSUE #36: Live switching not working
Doing a rebuild switch, will create the dirs but not create the bindfs mounts. 
Not sure if this is just a me issue because I use nixus, or if anyone else have this as well.

talyz:
Due to what is likely a bug in `bindfs` or `fuse`, the target path is sometimes missing from the mount entry. This causes false positives for the target directory having changed, leading to unnecessary remounts. Luckily, it seems that when this happens, the line instead contains the string `bindfs`, which it doesn't normally, so we can at least circumvent this issue to some degree.

Also, find the correct name of the activation script entry responsible for reloading systemd user services. The name was initially `reloadSystemD` but has been changed to `reloadSystemd`, causing failures due to the unmounts being done after the systemd services are reloaded.

Fixes #36

### DIFF
--- a/README.org
+++ b/README.org
@@ -92,17 +92,14 @@
             "Documents"
             "Videos"
             "VirtualBox VMs"
-            ".gnupg/private-keys-v1.d"
+            ".gnupg"
             ".ssh"
             ".nixops"
             ".local/share/keyrings"
             ".local/share/direnv"
           ];
           files = [
-            ".gnupg/pubring.kbx"
-            ".gnupg/sshcontrol"
-            ".gnupg/trustdb.gpg"
-            ".gnupg/random_seed"
+            ".screenrc"
           ];
         };
       }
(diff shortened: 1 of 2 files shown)


eyJhb:
This worked great! I could even do a switch from symlinks to the bindfs ones in one go! No rebooting or any weird stuff required.
Tried to add a new dir afterwards, and that worked as well.

LGTM!

lovesegfault (approved):

talyz:
@eyJhb Awesome!

The `bindfs` issue was fixed [upstream](https://github.com/mpartel/bindfs/issues/94)! I pushed a change that utilizes the fix if it's available. See https://github.com/NixOS/nixpkgs/pull/110727 for the package update.

lovesegfault (approved):

PR #39, opened 2021-01-24
home-manager: Add `allowOther` option to let other users read bind mounts

### LINKED ISSUE #35: Does not work with Docker
If you cd to some dir, that is managed with impermanence, and do

```
docker run --rm -it -v $(pwd):/src alpine
```

Then you will get a error like `docker: Error response from daemon: error while creating mount source path '/home/eyjhb/projects': mkdir /home/eyjhb/projects: file exists.`

This seems to be a problem with the bindfs method, as it worked just fine with symlinks (and works fine if I use the actual dir, not the one it bindfs).

talyz:
NixOS defaults to not letting fuse mounts be allowed to let other users read their contents. `bindfs` wants to give other users access and is therefore normally run with `--no-allow-other` to not throw an error.

Giving other users, mainly `root`, access to the bind mounts is, however, useful and works fine when
```nix
{
  programs.fuse.userAllowOther = true;
}
```

is declared in `configuration.nix`. This adds an option to choose whether to give other users access or not. It also prompts the user to set the `allowOther` attribute with a link to the documentation.

Should be merged after #38.

Fixes #35

### DIFF
--- a/home-manager.nix
+++ b/home-manager.nix
@@ -14,7 +14,7 @@ in
     home.persistence = mkOption {
       default = { };
       type = with types; attrsOf (
-        submodule {
+        submodule ({ name, ... }: {
           options =
             {
               directories = mkOption {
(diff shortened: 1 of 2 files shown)


home-manager.nix
```diff
@@ -14,25 +14,84 @@ in
...
+                '';
+              };
+
+              allowOther = mkOption {
+                type = with types; nullOr bool;
+                default = null;
+                example = true;
+                apply = x:
+                  if x == null then
+                    warn ''
```
cole-h, line 59: Is it necessary to display this warning? Why not just default to `false`?

talyz: It's not strictly necessary, no. It's mainly that I think most people would want to set it to `true`, but probably don't know it. Essentially, we can't set it to the sane default, since it clashes with the default NixOS configuration.

lovesegfault (approved):

lovesegfault (approved):

PR #44, opened 2021-08-29
nixos: Manage files using systemd services

### LINKED ISSUE #1: New API

talyz:
Manage files by creating a systemd oneshot service for each file. The service links or bind mounts the file as appropriate on start and removes the link or unmounts it when stopped. Whether a symlink or bind mount is used is determined by if the target exists - if it does, it's bind mounted, otherwise symlinked. To make sure files are available early enough, also run the start portion in the activation script.

This lifts the restriction on files being placed in `/etc` and should finally close #1.

### DIFF
--- a/nixos.nix
+++ b/nixos.nix
@@ -4,8 +4,23 @@ with lib;
 let
   cfg = config.environment.persistence;
   persistentStoragePaths = attrNames cfg;
+  mkMountScript = mountPoint: targetFile: ''
+    if [[ -L ${mountPoint} && $(readlink -f ${mountPoint}) == ${targetFile} ]]; then
+        echo "${mountPoint} already links to ${targetFile}, ignoring"
+    elif mount | grep -F ${mountPoint}' ' >/dev/null && ! mount | grep -F ${mountPoint}/ >/dev/null; then
+        echo "mount already exists at ${mountPoint}, ignoring"
+    elif [[ -e ${mountPoint} ]]; then
+        echo "A file already exists at ${mountPoint}!" >&2
+        exit 1
+    elif [[ -e ${targetFile} ]]; then
+        touch ${mountPoint}
+        mount -o bind ${targetFile} ${mountPoint}
+    else
+        ln -s ${targetFile} ${mountPoint}
+    fi
+  '';
 
-  inherit (pkgs.callPackage ./lib.nix { }) splitPath dirListToPath concatPaths;
+  inherit (pkgs.callPackage ./lib.nix { }) splitPath dirListToPath concatPaths sanitizeName;
 in
 {
   options = {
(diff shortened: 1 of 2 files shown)


colemickens:
I'm using this PR, and I'm seeing this behavior.

1. At some point, I get things all working.
2. I reboot.
3. Things come back correctly! (etc machine-id, and ssh host keys)
4. As soon as I try to activate a new generation I get errors:

```
...
updating GRUB 2 menu...
activating the configuration...
mount already exists at '/etc/machine-id', ignoring
reloading user units for cole...
setting up tmpfiles

```

Frustratingly there's a sort of silent rollback there and the new generation is not activated or marked as the running generation. I guess I can just reboot after I set the system profile every time, but that's not ideal...

nixos.nix
```diff
@@ -124,14 +154,51 @@ in
...
+          {
+            "bindOrLink-${sanitizeName targetFile}" = {
+              deps = [ "createDirsIn-${replaceStrings [ "/" "." " " ] [ "-" "" "" ] persistentStoragePath}" ];
+              text = ''
+                if [[ -L ${mountPoint} && $(readlink -f ${mountPoint}) == ${targetFile} ]]; then
+                    echo "${mountPoint} already links to ${targetFile}, ignoring"
+                    exit 0
+                elif mount | grep -F ${mountPoint}' ' >/dev/null && ! mount | grep -F ${mountPoint}/ >/dev/null; then
+                    echo "mount already exists at ${mountPoint}, ignoring"
+                    exit 0
```
talyz, line 180: @colemickens Thanks for testing! I think this is the culprit. It looks like a copy-paste-fail.

talyz: @colemickens Can you try it now?

colemickens:
Seems like it fixed it!

talyz:
Great! Thanks again for trying it out!

ambroisie:
What's the state on this PR? Is it going to be merged soon?

talyz:
I think it's done, but I would prefer to get some more testing and reviews before merging it.

Ping @etu @lovesegfault @cole-h

nixos.nix
```diff
@@ -124,14 +151,48 @@ in
...
+
+        bindOrLinkFileScript = persistentStoragePath: file:
+          let
+            targetFile = escapeShellArg (concatPaths [ persistentStoragePath file ]);
+            mountPoint = escapeShellArg file;
+          in
+          {
+            "bindOrLink-${sanitizeName targetFile}" = {
+              deps = [ "createDirsIn-${replaceStrings [ "/" "." " " ] [ "-" "" "" ] persistentStoragePath}" ];
+              text = ''
```
cole-h, line 171: I wonder if you {sh,c}ould abstract this out, since it looks like it's used here, and on line 77 (the script contents, at least). Otherwise, they risk falling out of sync and creating potentially difficult-to-debug issues.

cole-h: (In case it wasn't clear -- "this" being the script's text)

talyz: Yeah, true! I'll deduplicate this.

talyz: Done!

cole-h (approved):
Much better. I (still >.<) have no way to test this, but the diff looks sane to me.

lovesegfault:
I'm going to test this tomorrow without fault.

lovesegfault:
Seems to be working AOK :)

* I wish we had separate services for binding and linking, so `bind-foo`, `link-bar`, instead of `bindOrLink-foo`, `bindOrLink-bar`
* I wonder whether the unit names having `impermanence` on them would be good? This way a user knows what they are when looking at `systemctl`

ambroisie:
@lovesegfault I think having `impermanence-` as a prefix is necessary to be able to distinguish between `link-foo` and `bind-bar` while still having a simple "namespace".

talyz:
@lovesegfault Well, the problem is that we don't know wether we'll link or bind mount until the service runs, so I think `bindOrLink` is as accurate as it gets. We could of course append `impermanence` or maybe `persistence`, since that's what the option is called. It would make the names very long, but maybe that's okay.

lovesegfault:
> @lovesegfault Well, the problem is that we don't know wether we'll link or bind mount until the service runs, so I think `bindOrLink` is as accurate as it gets. We could of course append `impermanence` or maybe `persistence`, since that's what the option is called. It would make the names very long, but maybe that's okay.

How about just `persist-$path`?

talyz:
Yeah, that sounds good!

talyz:
@lovesegfault Changed! :)

nixos.nix
```diff
@@ -57,29 +72,44 @@ in
...
-            "${replaceStrings [ "/" "." " " ] [ "-" "" "" ] file}"
-            { }
-            "ln -s '${file}' $out";
-
-        # Create environment.etc link entry.
-        mkLinkNameValuePair = persistentStoragePath: file: {
-          name = removePrefix "/etc/" file;
-          value = { source = link (concatPaths [ persistentStoragePath file ]); };
-        };
+        mkBindOrLinkFileService = persistentStoragePath: file:
```
cole-h, line 77: Nit, but I think all references to `BindOrLink` should be replaced with `Persist` to align with the new prefix.

nixos.nix
```diff
@@ -124,14 +154,34 @@ in
...
+          {
+            "createDirsIn-${replaceStrings [ "/" "." " " ] [ "-" "" "" ] persistentStoragePath}" =
+              noDepEntry (concatMapStrings
+                (mkDirWithPerms persistentStoragePath)
+                (directories ++ files));
+          };
+
+        dirCreationScripts = foldl' recursiveUpdate { } (map mkDirCreationScriptForPath persistentStoragePaths);
+
+        bindOrLinkFileScript = persistentStoragePath: file:
```
cole-h, line 166: Here too.

nixos.nix
```diff
@@ -124,14 +154,34 @@ in
...
+          };
+
+        mkBindOrLinkFileScripts = persistentStoragePath:
+          foldl' recursiveUpdate { } (map (bindOrLinkFileScript persistentStoragePath) cfg.${persistentStoragePath}.files);
+
+        bindOrLinkFileScripts =
+          foldl' recursiveUpdate { } (map mkBindOrLinkFileScripts persistentStoragePaths);
       in
-      listToAttrs (map mkDirCreationScriptForPath persistentStoragePaths);
+      dirCreationScripts // bindOrLinkFileScripts;
```
cole-h, line 184: and here.

talyz:
@cole-h All done! I kept the names of the `ExecStart` and `ExecStop` scripts, since I found the names more descriptive than something like `persist-start-*`, but all the other instances were changed.

lovesegfault:
Thank you for tackling this @talyz!

PR #61, opened 2021-12-27
home-manager: allow dependencies on bind mounts

### LINKED ISSUE #60: home-manager: race condition with certain services
I use `syncthing` via home-manager, so I have to bind-mount its mutable directories. The problem is that both the bind-mount and `syncthing` itself are started via systemd, so `syncthing` may start and recreate its configuration from scratch before `bindfs` gets a chance to do its thing.

My current solution is to turn the `bindfs` unit into a `forking`-type systemd unit and make the `syncthing` unit depend on it. Is there a known better solution? If not I will PR my patch.

lourkeur:
Closes #60


talyz:
Looks good! I tried stopping units and the processes seems to be terminated correctly.

PR #62, opened 2021-12-28
home-manager: Start the bind units as early as possible

talyz:
Make sure the bind mount units start as early as possible in the user session. This should hopefully make issues like #60 less prevalent.


lourkeur (approved):
After booting on it twice, it like it does prevent the race condition. It also makes sense on paper because `syncthing.service` comes `After` `basic.target`. LGTM

PR #65, opened 2021-12-31
nixos: types.str -> types.path

winterqt:


winterqt:
Hm, not sure if this is a good idea after thinking about it further. I'll leave it open just in case someone thinks differently.

adisbladis:
I think path's are more "dangerous" as they are easier to mishandle than a string, where the mishandled case leads to copying the files to the store.

winterqt:
@adisbladis Yeah, that was my reasoning for reconsidering this change. Strings are much harder to misuse in this case.

The main reason this change was proposed in the first place was that it provided a way to check for valid path syntax. How would you feel about changing this to add a check to the string types to check for things such as not being empty, requiring a leading slash, and not having a trailing slash?

adisbladis:
> How would you feel about changing this to add a check to the string types to check for things such as not being empty, requiring a leading slash, and not having a trailing slash?

That sounds very welcome to me!

talyz:
This would be a bit more complicated with #70 in place, where paths in `users` shouldn't start with a slash, so user paths would have to have a different type/check. Definitely doable, though.

Issue #22, opened 2020-11-09
Mention that filesystems are mounted alphabetically


Nadrieril:
I just switched my root to be on tmpfs. I decided to start with mounting the real root fs to `/persistent` and bind-mounting everything from there. Except that it wouldn't boot because it tried to mount `/etc` before mounting `/persistent`.
I discovered that naming the mount point `/apersistent` fixed it tho x) Apparently the filesystems are loaded alphabetically. Is that expected or did I miss something? Will this stay that way with future nix versions? If that's expected behavior then I think it deserves to be mentioned in the README.

talyz:
Yes, you need to mark the persistent filesystem `neededForBoot` in order for it to be mounted early enough. This is mentioned in the `README.org` at the end of the `NixOS` section (on line 61).

Nadrieril:
I did that, but it was not enough, because `/nix` is also neededForBoot, but /nix is supposed to be bind-mounted from /persistent.

lovesegfault:
In the past I think I worked around this by making my persistent fs be `/nix` and make `/nix/state` my impermanence dir. You can see the setup here: https://github.com/lovesegfault/nix-config/blob/master/systems/fourier.nix

It's not great, but it works.

talyz:
I see. Is there a reason to set it up like this, if you don't mind me asking? If you want to use `/nix` for persistent storage, you can just use a subpath of it, like `/nix/persistent`.

I also use `/persistent` for persistent storage, but it's a mounted btrfs subvolume, not a bind mount. Here is my filesystem config: https://github.com/talyz/nixos-config/blob/master/machines/evals/configuration.nix#L82-L110. I also create a fresh btrfs subvolume on boot and use it as root, which saves memory and provides a way to recover in case of crashes, power outages or forgetfulness ;)

Nadrieril:
My main reason for setting up `/persistent` the way I do is that I started from an existing normal nixos installation. So `/persistent` points to the root of a normal installation, with its /nix, /home, /etc and so on. Btw I want to mention that apart from having to name the mountpoint `/apersistent`, the rest of the transition was extremely smooth!
Fresh btrfs sounds like a pretty cool idea! But so far my non-persisted `/` takes 20M tops.

talyz:
[...]

PR #66, opened 2022-01-03
nixos: Mount binds after the persistent storage path

olmokramer:
This sets the `depends` (introduced in https://github.com/NixOS/nixpkgs/pull/86967) of each bind mount so that it will only be mounted after the persistent storage path has been mounted, as suggested in https://github.com/nix-community/impermanence/issues/22#issuecomment-847879359.

An obvious issue with this is that this will break older NixOS where the option does not exist yet. Any suggestions on how to work around that?

### DIFF
--- a/nixos.nix
+++ b/nixos.nix
@@ -120,6 +120,7 @@ in
             device = concatPaths [ persistentStoragePath dir ];
             noCheck = true;
             options = [ "bind" ];
+            depends = [ persistentStoragePath ];
           };
         };
 


cole-h:
If GitHub's tags are to be believed, that PR is present in both 21.05 and 21.11, and support for any version of NixOS older than those have expired. IME, this is fine; however, I'd want to document this new version bound. Something like "as of PR #66 (with a link), NixOS 21.05 or later is required for proper functionality", or however you'd care to word it with a similar effect.

Though, this is also just my opinion. cc @talyz @lovesegfault @etu

olmokramer:
Ah I did not know that the patch was backported to all supported versions of NixOS. I agree that the change in version requirements should be documented as @cole-h suggests. I'll try to find some time in the weekend to make the changes.

talyz:
Fixed in #115

olmokramer:
Oh hey I totally forgot about this one. Thanks for picking it up again @Kranzes!

PR #70, opened 2022-01-16
nixos: Refactor and implement support for user files + directories

talyz:
An overhaul of the current NixOS module which adds a `users` feature, makes the module more extensible, adds permissions handling, and asserts that no duplicates are specified:

## Implement support for user files and directories

Allow user files and directories to be specified as follows:

```nix
{
    users.talyz = {
      files = [
        ".screenrc"
      ];
      directories = [
        "Downloads"
      ];
    };
  };
}
```

This provides an alternative to the home-manager module and may even deprecate it in the future.

## Refactor the module to simplify future extensibility

Implement support for a submodule representation for files and directories. Strings are automatically converted to appropriate submodule representations and each file and directory is handled based only on their respective submodule's attributes. This means that for most files, a string will suffice, but if more advanced options need to be set for the specific files or directories, a submodule can be used instead. It also, arguably, simplifies the implementation a bit.

## Add support for custom permissions on created directories

Building on the extensibility mentioned above, allow the owner and mode to be set when directories are created in persistent storage by the `create-directories.bash` script. Very useful for directories used to store secrets.

Also, make sure the user directories are created with reasonable defaults, i.e. owned by the user and its group, not by `root:root`.

Here is an example of this in use:

```nix
{
    directories = [
      "/var/log"
      "/etc/NetworkManager/system-connections"
      { directory = "/var/lib/colord"; user = "colord"; group = "colord"; mode = "u=rwx,g=rx,o="; }
    ];
    files = [
      "/etc/machine-id"
      { file = "/etc/nix/id_rsa"; parentDirectory = { mode = "u=rwx,g=,o="; }; }
    ];
    users.talyz = {
      directories = [
        "Downloads"
        "Music"
        { directory = ".gnupg"; mode = "0700"; }
        { directory = ".ssh"; mode = "0700"; }
        { directory = ".nixops"; mode = "0700"; }
        { directory = ".local/share/keyrings"; mode = "0700"; }
        ".local/share/direnv"
      ];
    };
  };
}
```

## Assert that no duplicate files or directories are specified

Make sure the same files or directories aren't specified multiple times, either between persistent storage paths or in the same path.

### DIFF
--- a/README.org
+++ b/README.org
@@ -40,10 +40,11 @@
             "/var/lib/bluetooth"
             "/var/lib/systemd/coredump"
             "/etc/NetworkManager/system-connections"
+            { directory = "/var/lib/colord"; user = "colord"; group = "colord"; mode = "u=rwx,g=rx,o="; }
           ];
           files = [
             "/etc/machine-id"
-            "/etc/nix/id_rsa"
+            { file = "/etc/nix/id_rsa"; parentDirectory = { mode = "u=rwx,g=,o="; }; }
           ];
           users.talyz = {
             directories = [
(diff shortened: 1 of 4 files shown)


etu (approved):
Works great! I'm using this branch for now at least on one system :)

lovesegfault (approved):
Let's get this landed :D

lovesegfault:
Ah, this causes an eval failure if you import the module put don't use it: 

https://github.com/lovesegfault/nix-config/runs/5011183485?check_suite_focus=true

PR #73, opened 2022-01-31
fix(nixos): allow eval to succeed when the module is unused

lovesegfault:
Fixes an issue from #70 where system configs won't eval if no
`environment.persist` confg exists.

~I think this fix is a little hacky, there's probably a better/clear way of
handling this.~ Thought of a better way.

### DIFF
--- a/nixos.nix
+++ b/nixos.nix
@@ -12,7 +12,8 @@ let
 
   cfg = config.environment.persistence;
   users = config.users.users;
-  allPersistentStoragePaths = zipAttrsWith (_name: flatten) (attrValues cfg);
+  allPersistentStoragePaths = { directories = [ ]; files = [ ]; users = [ ]; }
+    // (zipAttrsWith (_name: flatten) (attrValues cfg));
   inherit (allPersistentStoragePaths) files directories;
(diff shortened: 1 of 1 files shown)


nixos.nix
```diff
@@ -13,7 +13,8 @@ let
   cfg = config.environment.persistence;
   users = config.users.users;
   allPersistentStoragePaths = zipAttrsWith (_name: flatten) (attrValues cfg);
```
lovesegfault, line 15: I imagine the way we actually want to fix this is to change this call so that it produces the empty attrs we want.

lovesegfault: Something like this perhaps?
```Nix
{ files = [ ]; directories = [ ]; users = [ ]; } // (zipAttrsWith (_name: flatten) (attrValues cfg))
```

lovesegfault: Yeah, I like this better

lovesegfault:
@talyz I'm going to fast-merge this to un-break master since the patch is so small, we can re-hash it in a follow-up if you come up with a nicer fix?

talyz:
I think this looks good! Maybe there's a nicer fix, but it would likely require a larger change. Thanks for fixing the issue!

Issue #74, opened 2022-02-01
[Bug] Created user directories is owned by root, and not the user


a12l:
My `$HOME` is rolled back to a blank ZFS snapshot at every startup. The dotfiles that I want to keep between reboots is listed in Impermanence's directory and file lists; and the other files that I want to keep is located inside ZFS datasets that gets mounted in `$HOME`.

My problem using the new way to list users' directories and files is that the parent directories is created and owned by `root`. This cause a lot of problems. For example, if I have this in my NixOS configuration

```
environment.persistence."/persistent".users.a12l.directories = [ ".local/share/fonts" ];
```

then I get this when I log in

```
$ ls -la|rg .local
drwxr-xr-x  4 root root     4 Feb  1 21:33 .local

$ ls -lR .local
.local:
total 33
drwxr-xr-x 3 root root 3 Feb  1 21:33 share

.local/share:
total 17
drwxr-xr-x 3 a12l users 9 Jan 18 15:39 fonts

.local/share/fonts:
total 852
-rw-r--r-- 1 a12l users  44732 Jan 18 15:31 all-the-icons.ttf
-rw-r--r-- 1 a12l users 489672 Jan 18 15:31 file-icons.ttf
-rw-r--r-- 1 a12l users 152796 Jan 18 15:31 fontawesome.ttf
-rw-r--r-- 1 a12l users 128180 Jan 18 15:31 material-design-icons.ttf
-rw-r--r-- 1 a12l users  52544 Jan 18 15:31 octicons.ttf
drwxr-xr-x 2 a12l users     18 Jan 30 03:37 p
-rw-r--r-- 1 a12l users  99564 Jan 18 15:31 weathericons.ttf

.local/share/fonts/p:
total 19744
-rw-r--r-- 1 a12l users 2379348 Jan 30 03:26 PragmataProB_0829.ttf
-rw-r--r-- 1 a12l users 2404140 Jan 30 03:26 PragmataProB_liga_0829.ttf
-rw-r--r-- 1 a12l users 2262112 Jan 30 03:26 PragmataProI_0829.ttf
-rw-r--r-- 1 a12l users 2287048 Jan 30 03:26 PragmataProI_liga_0829.ttf
-rw-r--r-- 1 a12l users 1326160 Jan 30 03:26 PragmataPro_Mono_B_0829.ttf
-rw-r--r-- 1 a12l users 1351116 Jan 30 03:26 PragmataPro_Mono_B_liga_0829.ttf
[...]

PR #75, opened 2022-02-02
nixos: Fix the path splitting in create-directories.bash

talyz:
Paths are currently split on bash's default delimiters (space, tab and newline) with slashes converted to space. This means paths containing any such delimiters are incorrectly handled, as @etu reported here: https://github.com/nix-community/impermanence/issues/74#issuecomment-1027643883.

Fix this by temporarily replacing bash's default delimiters with `/`, making sure this is the only character we split on.


etu (approved):
Works fine here :)

PR #76, opened 2022-02-04
feat(nixos): Support Hiding of Bind Mounts

### LINKED ISSUE #64: (home-manager) File manager (Nautilus) shows persistent folders as ejectable drives
Nautilus shows my Documents and Downloads folder as ejectable drive as well as showing drive icons beside them in the file view in /home. A solution I am thinking of would be to allow setting mount options with impermanence so the x-gvfs-hide option can be used.

Unsure whether this option would be set per directory/file or per persistent mount (set mount option that affects all of /nix/persist for example).

maydayv7:
This is done by exposing an option `hideMounts` to set the `x-gvfs-hide` mount option on all the bind mounts

Fixes #64

### DIFF
--- a/README.org
+++ b/README.org
@@ -35,6 +35,7 @@
     #+begin_src nix
       {
         environment.persistence."/persistent" = {
+          hideMounts = true;
           directories = [
             "/var/log"
             "/var/lib/bluetooth"
@@ -136,6 +137,11 @@
 
       If the file exists in persistent storage, it will be bind
       mounted to the target path; otherwise it will be symlinked.
+
+    - ~hideMounts~ allows you to specify whether to hide the
+      bind mounts from showing up as mounted drives in the file
+      manager. If enabled, it sets the mount option ~x-gvfs-hide~
+      on all the bind mounts.
 
     - ~users.talyz~ handles files and directories in ~talyz~'s home
       directory
--- a/nixos.nix
+++ b/nixos.nix
@@ -5,7 +5,7 @@ let
     types foldl' unique noDepEntry concatMapStrings listToAttrs
     escapeShellArg escapeShellArgs replaceStrings recursiveUpdate all
     filter filterAttrs concatStringsSep concatMapStringsSep isString
-    catAttrs;
+    catAttrs optional;
 
   inherit (pkgs.callPackage ./lib.nix { }) splitPath dirListToPath
     concatPaths sanitizeName duplicates;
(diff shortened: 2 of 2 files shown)


README.org
```diff
@@ -137,6 +138,11 @@
       If the file exists in persistent storage, it will be bind
       mounted to the target path; otherwise it will be symlinked.
 
+    - ~hideMounts~ allows you to specify whether to hide the
```
cole-h, line 141: Instead of an option that only affects GUI file managers, maybe it should instead be possible to modify the mount options wholesale?

nixos.nix
```diff
@@ -332,7 +341,10 @@ in
           value = {
             device = concatPaths [ persistentStoragePath directory ];
             noCheck = true;
-            options = [ "bind" ];
+            options = [ "bind" ] ++ (if cfg.${persistentStoragePath}.hideMounts then
```
cole-h, line 344: This can be more concisely written with `optional` or `optionals` -- e.g. `++ optional cfg.${persistentStoragePath}.hideMounts "x-gvfs-hide";` or `++ optionals .... [ "x-gvfs-hide" ];`

README.org
maydayv7: That would change the scope of this PR, but yes, an option to specify mount options seems more useful

lovesegfault: I think Cole is asking whether there is a mount option that applies to things other than GUI file managers. So that, for example, when I do `lsblk` I don't see all the bind mounts.

I don't think such an option exists though.

cole-h: No, I was asking to make this more general :P

maydayv7: Would it be fine if there was a `hideMounts` option as well as another option to specify additional mount options?

cole-h: I don't particularly care which way this goes (mostly because I don't actually use this (........yet.......)), to be honest. I just figured it'd be nicer if it was more generic instead of hardcoding support for something that "only" is noticeable in GUI file managers.

So, aside from this, I think this looks good to me :)

maydayv7: Yep, understood, the main reason I wanted to keep the option is because `x-gvfs-hide` is not an obvious mount flag and it would be much more convenient to include it as a toggleable option (but it would also be fine with a generic mounts option with some documentation on useful ones)

lovesegfault:
Deferring adding a more general `extraMountFlags` in another PR, if people feel the need for it.

PR #82, opened 2022-02-11
Add default NixOS module to flake

crunchcat:
This adds a “default” NixOS module in `flake.nix`. This makes using this in flakes a little less verbose and repetitive.

Before this change:

```nix
nixpkgs.lib.nixosSystem {
  modules = [
    ./configuration.nix
    impermanence.nixosModules.impermanence
  ];
}
```

After this change:

```nix
nixpkgs.lib.nixosSystem {
  modules = [
    ./configuration.nix
    impermanence.nixosModule
  ];
}
```


talyz:
Thanks! :)

PR #83, opened 2022-02-11
nixos: Fix bind mounts in VM build

### LINKED ISSUE #50: impermanence doesn't work with `qemu-vm.nix built systems.
I'm using nixos testing machinery to setup several nixos VMs for ad-hoc testing of a distributed service. I want to persist some state between run (in particular ssh host keys and tailscale configuration). I tried using impermanence, but it looks like it doesn't work, since `qemu-vm.nix` overrides `config.fileSystems` set by impermanence.

devplayer0:
Fixes #50. By exposing the bind mount `fileSystems` entries in `virtualisation.fileSystems` as well, the VM module will include them in its overridden `fileSystems`.

Note that the `virtualisation.fileSystems` option needs to be declared as a dummy since this option won't exist unless the VM module is actually imported.

### DIFF
--- a/nixos.nix
+++ b/nixos.nix
@@ -19,6 +19,21 @@ let
     cp ${./mount-file.bash} $out
     patchShebangs $out
   '';
+
+  # Create fileSystems bind mount entry.
+  mkBindMountNameValuePair = { directory, persistentStoragePath, ... }: {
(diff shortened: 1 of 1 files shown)


talyz:
This seems like a reasonable fix. Thank you!

PR #84, opened 2022-02-14
nixos: Don't exit with an error in the activation script

### LINKED ISSUE #77: Kernel panic resulting from a possible precedence issue
Hi!

I was previously using impermanence pinned to commit 58558845bc68dcf2bb32caa80564f7fe3f6cbc61 and, having seen the recent improvements to the module, decided to update to commit ff540dbe20556f6119d80f5c79796a0698a4ee38.

Unfortunately, somewhere in between, possibly due to the addition of #44, I believe that the `/etc/NIXOS` file, which I currently have configured to be persisted, as it is stated to be necessary for NixOS to recognize its own root partition, became involved in a precedence issue, where it causes NixOS to panic due to its potentially premature loading. The following screenshot exemplifies this occurrence (sorry for the poor quality but, as I do not have `systemctl` logs persisted, I am unable to directly copy the relevant log lines):

image

What occurs after the `/etc/NIXOS` clash is that systemd is not found and the boot process is aborted.

I apologize for the vagueness of my explanation, but I am at a loss as to what else may be the root cause of this issue, as well as any potential workarounds.

talyz:
Exiting with an error in the activation script causes the whole script to fail and system boot thus fails with a kernel panic. Since the point of exiting with an error is mainly to make the systemd service fail, let's do it only when the script runs as a service.

Fixes #77

### DIFF
--- a/nixos.nix
+++ b/nixos.nix
@@ -5,7 +5,7 @@ let
     types foldl' unique noDepEntry concatMapStrings listToAttrs
     escapeShellArg escapeShellArgs replaceStrings recursiveUpdate all
     filter filterAttrs concatStringsSep concatMapStringsSep isString
-    catAttrs optional literalExpression;
+    catAttrs optional optionalString literalExpression;
 
   inherit (pkgs.callPackage ./lib.nix { }) splitPath dirListToPath
     concatPaths sanitizeName duplicates;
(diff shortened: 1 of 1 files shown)


talyz:
Superseded by #85

PR #88, opened 2022-03-26
Add `force` option for replacing existing files

lheckemann:
I wrote this in hopes that I could manage `/etc/shadow` with impermanence. Unfortunately, this does not play along with the `rename`-based atomic update mechanism applied by both `passwd` and nixos's `update-users-groups.pl`, so I don't have a good example to use in the documentation...

### DIFF
--- a/README.org
+++ b/README.org
@@ -46,6 +46,7 @@
           files = [
             "/etc/machine-id"
             { file = "/etc/nix/id_rsa"; parentDirectory = { mode = "u=rwx,g=,o="; }; }
+            { file = "/etc/shadow"; force = true; }; }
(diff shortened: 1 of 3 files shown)


KiaraGrouwstra:
as others linked here might wonder as well, what might be alternatives for `shadow` (/ `passwd`)?
perhaps as a bonus question: is there a way one might check if such a particular file should work with this, or fall into a similar category as those exceptions?

mibmo:
how come this hasn't been merged? just due to a lack of documentation example?

talyz:
It's a draft currently lacking a use-case and a working example.

PR #99, opened 2022-07-14
home-manager: adds configurable symlink/bindfs option

### LINKED ISSUE #42: Why a bindfs rather than a symlink?
Just wondered why we bother with bindfs rather than just using a symlink? Wouldn't a bindfs allow blocking until the persistent filesystem is mounted, and wouldn't it not have to wait until half way through the user's session start to be able to be used?

I'm just seeing a lot of things that aren't yet mounted on login even though I should see every mount first, e.g. desktop settings, wifi password keychains etc

Misterio77:
Hello, all! A huge thanks to the maintainers :), I've been using impermanence for a while and really love it.

---

As mentioned in https://github.com/nix-community/impermanence/issues/42, bindfs causes some issues with a few specific programs (steam, for example). I'm aware that we have options for using the NixOS module to avoid bindfs, but that may not be preferred (or possible) for some.

This PR allows for setting a specific `method` for each individual directory. They can (atm) be either "bindfs" (the default if not specified) or "symlink".

I've reused the code that links files, but now it runs on `directories` that have `method = "symlink"` as well.

I made sure that changing between the two methods works reliably. For that end, I've added:
- a `cleanEmptyLinkTargets` activation script that handles changing from bindfs to symlink. If a link's intended location has something mounted, it is unmounted (earlier than usual) and its mount point is deleted (safely with `rm -d`).
- a `mount | grep` condition on the usual unmount phase, to avoid systemd trying to unmount a symlink (causes error messagens and a few seconds of hanging, although it ultimately works).

~Oh, and by the way, git got a little confused with the repeated unmount code and messed up the diff. The only change I made to `mkUnmount` was adding the `mount | grep` check.~

As we have three places with the same unmount code now, I've factored it out to a function. Diff looks fine now.

Hope this is useful, and please let me know if there's anything to be improved!

---
Fixes #42

### DIFF
--- a/home-manager.nix
+++ b/home-manager.nix
@@ -6,7 +6,32 @@ let
 
   persistentStoragePaths = attrNames cfg;
 
+  getDirPath = v: if isString v then v else v.directory;
+  getDirMethod = v: v.method or "bindfs";
+  isBindfs = v: (getDirMethod v) == "bindfs";
+  isSymlink = v: (getDirMethod v) == "symlink";
+
   inherit (pkgs.callPackage ./lib.nix { }) splitPath dirListToPath concatPaths sanitizeName;
+
+  mount = "${pkgs.util-linux}/bin/mount";
+  unmountScript = mountPoint: tries: sleep: ''
+    triesLeft=${toString tries}
+    if ${mount} | grep -F ${mountPoint}' ' >/dev/null; then
+        while (( triesLeft > 0 )); do
+            if fusermount -u ${mountPoint}; then
+                break
+            else
+                (( triesLeft-- ))
+                if (( triesLeft == 0 )); then
+                    echo "Couldn't perform regular unmount of ${mountPoint}. Attempting lazy unmount."
+                    fusermount -uz ${mountPoint}
+                else
+                    sleep ${toString sleep}
+                fi
+            fi
+        done
+    fi
+  '';
 in
 {
   options = {
@@ -18,7 +43,25 @@ in
           options =
             {
               directories = mkOption {
-                type = with types; listOf str;
+                type = with types; listOf (either str (submodule {
+                  options = {
+                    directory = mkOption {
+                      type = str;
+                      default = null;
+                      description = "The directory path to be linked.";
+                    };
+                    method = mkOption {
+                      type = types.enum [ "bindfs" "symlink" ];
+                      default = "bindfs";
+                      description = ''
+                        The linking method that should be used for this
+                        directory. bindfs is the default and works for most use
+                        cases, however some programs may behave better with
+                        symlinks.
+                      '';
+                    };
+                  };
+                }));
                 default = [ ];
                 example = [
                   "Downloads"
@@ -31,10 +74,15 @@ in
                   ".ssh"
                   ".local/share/keyrings"
                   ".local/share/direnv"
+                  {
+                    directory = ".local/share/Steam";
+                    method = "symlink";
+                  }
                 ];
                 description = ''
                   A list of directories in your home directory that
-                  you want to link to persistent storage.
+                  you want to link to persistent storage. You may optionally
+                  specify the linking method each directory should use.
                 '';
               };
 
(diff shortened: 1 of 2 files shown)


home-manager.nix
```diff
@@ -6,6 +6,13 @@ let
 
   persistentStoragePaths = attrNames cfg;
 
+  getPath = v: if isString v then v else v.path;
+  getMethod = v: v.method or "bindfs";
+  isBindfs = v: (getMethod v) == "bindfs";
+  isSymlink = v: (getMethod v) == "symlink";
```
Misterio77, line 12: Some convenience

home-manager.nix
```diff
@@ -6,6 +6,13 @@ let
...
+
+  mount = "${pkgs.util-linux}/bin/mount";
```
Misterio77, line 14: Moved to top level as a couple places use it now

home-manager.nix
Misterio77, line 12: Some convenience.

home-manager.nix
```diff
@@ -188,19 +221,21 @@ in
...
               triesLeft=6
-              while (( triesLeft > 0 )); do
-                  if fusermount -u ${mountPoint}; then
-                      exit 0
-                  else
-                      (( triesLeft-- ))
-                      if (( triesLeft == 0 )); then
-                          echo "Couldn't perform regular unmount of ${mountPoint}. Attempting lazy unmount."
-                          fusermount -uz ${mountPoint}
+              if ${mount} | grep -F ${mountPoint}' ' >/dev/null; then
```
Misterio77, line 224: This `if` is this section's actual change.

home-manager.nix
```diff
@@ -251,6 +286,7 @@ in
     home.activation =
       let
         dag = config.lib.dag;
+        mount = "${pkgs.util-linux}/bin/mount";
```
Misterio77, line 289: Moved it up as there's a few different places relying on it now.

home-manager.nix
```diff
@@ -317,6 +352,43 @@ in
           in
           ''
             if [[ -n ''${mountedPaths[${mountPoint}]+x} ]]; then
+                triesLeft=3
+                # Check to avoid trying to unmount a symlink
+                if ${mount} | grep -F ${mountPoint}' ' >/dev/null; then
```
Misterio77, line 357: Git messed up the diff, but the addition of this `if` is the actual change to `mkUnmount`.

home-manager.nix
```diff
@@ -332,16 +404,31 @@ in
                     fi
                 done
             fi
+
```
Misterio77, line 407: Can we somehow make the above section DRYier? Not a big fan of copying and pasting the unmount stuff (although it's already duplicated on both `mkUnmount` and `mkBindMountService`.

home-manager.nix
```diff
@@ -332,16 +404,31 @@ in
...
           '';
 
-        mkUnmountsForPath = persistentStoragePath:
+        mkLinkCleanupForPath = persistentStoragePath:
           concatMapStrings
-            (mkUnmount persistentStoragePath)
-            cfg.${persistentStoragePath}.directories;
+            (mkLinkCleanup persistentStoragePath)
+            (map getPath (filter isSymlink cfg.${persistentStoragePath}.directories));
+
```
Misterio77, line 419: Go home, git. You're drunk.

For the record, I didn't actually remove `mkUnmountsForPath` (it's on line 375 now), git decided to diff it that way.

talyz (changes requested):
Diff looks good and it works as expected for me. Thanks a lot for doing this! Just one tiny thing: could you rename `path` to `directory` to match the NixOS module?

Misterio77:
No problem at all! Thank you for this awesome tool :)

> could you `rename` path to `directory`

Done!

Misterio77:
Oops removed the review request by mistake, sorry

talyz:
It still needs to be updated in the readme. Also, please squash the update into the previous commit.

Misterio77:
All done :)

talyz:
Great work! Thank you!

PR #100, opened 2022-07-17
Make bash scripts work in cross-compiled environments.

plietar:
When cross-compiling, `patchShebangs` requires the host platform's bash
to be present in the HOST_PATH environment variable. However, when
using `pkgs.runCommand`, only the build platform's bash is added to the
PATH. The result is that the shebang is not replaced, and the script
fails to run because the activation scripts don't have `bash` in their
environment.

Using an explicit `mkDerivation` lets us add the host bash as a
`buildInput`, which makes `patchShebangs` work as expected.

For context, I'm using impermanence on a Raspberry Pi 0. Since no
binary cache exists for that platform, I cross-compile the entire OS and
configuration from an x86_64 host and only push the resulting closure.

### DIFF
--- a/nixos.nix
+++ b/nixos.nix
@@ -15,7 +15,7 @@ let
   allPersistentStoragePaths = { directories = [ ]; files = [ ]; users = [ ]; }
     // (zipAttrsWith (_name: flatten) (attrValues cfg));
   inherit (allPersistentStoragePaths) files directories;
-  mountFile = pkgs.runCommand "impermanence-mount-file" { } ''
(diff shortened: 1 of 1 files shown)


talyz (changes requested):
Nice catch! It should also be possible to keep `runCommand`in place and just add `buildInputs` to its argument set to get an even smaller change. Could you try that? :)

plietar:
I'd completely missed the fact that `runCommand` accepts the usual derivation attributes.
Should be good now.

talyz:
Thanks!

PR #106, opened 2022-11-05
Remove /etc/nix/id_rsa from example

jakobkukla:
I couldn't find any mention of this file except sources related with this repository. Furthermore the example changes the permissions of the `/etc/nix` directory. This breaks the new nix command for non-root users because the nix config file isn't readable anymore.

### DIFF
--- a/README.org
+++ b/README.org
@@ -45,7 +45,6 @@
           ];
           files = [
             "/etc/machine-id"
-            { file = "/etc/nix/id_rsa"; parentDirectory = { mode = "u=rwx,g=,o="; }; }
           ];
           users.talyz = {
             directories = [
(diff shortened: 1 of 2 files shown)


talyz:
The purpose of this file is just to serve as an example of how to use the options, not as a default config. It's the ssh key I use for remote building in my setup. I'd prefer to replace this example with something else if it's confusing.

jakobkukla:
I was aware that this only served as an usage example. Still it's rather confusing since it's next to sensible defaults like `/etc/NetworkManager/system-connections` and `/etc/machine-id`. How about we use something like `/etc/ssh/ssh_known_hosts` instead, to avoid that users who blindly copy the example end up with a broken nix command?

talyz:
AFAICT, the global known hosts file would mainly be used for declared known hosts and shouldn't need to be persisted, so it wouldn't be a good example either. Maybe we could use the ssh host keys as examples?

jakobkukla:
Closing as #151 fixed this.

PR #109, opened 2022-11-13
nixos: Improve directory creation and permission and ownership assignment

talyz:
Construct directory items for all parent directories of the user specified files and directories, assigning better default permissions
and ownership to each and removing this responsibility from the create-directories script.

This means that all parent directories of root directories will now have the default permissions and ownership, not inherit them from the child. User directories are assigned default user ownership. The home directory itself is handled specially to make sure it is owned by the user, not readable by anyone else and its parent gets default root ownership.

To illustrate this with an example, here is a directory specification and the ownership and permissions that could potentially be assigned to the parent directories, given none of them yet exist in persistent storage:

```nix
{
    users.talyz = {
      directories = [
        { directory = ".local/share/secret"; mode = "0500"; }
      ];
    };
  };
}
```

#### Before
```
/home                            talyz:talyz   0500
/home/talyz                      talyz:talyz   0500
/home/talyz/.local               talyz:talyz   0500
/home/talyz/.local/share         talyz:talyz   0500
/home/talyz/.local/share/secret  talyz:talyz   0500
```

#### After
```
/home                            root:root     0755
/home/talyz                      talyz:talyz   0700
/home/talyz/.local               talyz:talyz   0755
/home/talyz/.local/share         talyz:talyz   0755
/home/talyz/.local/share/secret  talyz:talyz   0500
```

Also:
- use the `coercedTo` type to convert from strings to file / directory items
- change the internal semantics of `file` and `directory`, introducing the `filePath`, `directoryPath` and `home` internal options
- formatting fixes

cc @tomeon

### DIFF
--- a/nixos.nix
+++ b/nixos.nix
@@ -205,7 +205,7 @@ in
                                 '';
                               };
                               files = mkOption {
-                                type = listOf (either str userFile);
+                                type = listOf (coercedTo str (f: { file = f; }) userFile);
                                 default = [ ];
                                 example = [
                                   ".screenrc"
@@ -214,22 +214,10 @@ in
                                   Files that should be stored in
                                   persistent storage.
                                 '';
-                                apply =
-                                  map (file:
-                                    if isString file then
-                                      {
-                                        inherit persistentStoragePath;
-                                        file = concatPaths [ config.home file ];
-                                        parentDirectory = userDefaultPerms;
-                                      }
-                                    else
-                                      file // {
-                                        file = concatPaths [ config.home file.file ];
-                                      });
                               };
 
                               directories = mkOption {
-                                type = listOf (either str userDir);
+                                type = listOf (coercedTo str (d: { directory = d; }) userDir);
                                 default = [ ];
                                 example = [
                                   "Downloads"
@@ -242,17 +230,6 @@ in
                                   Directories to bind mount to
                                   persistent storage.
                                 '';
-                                apply =
-                                  map (directory:
-                                    if isString directory then
-                                      userDefaultPerms // {
-                                        directory = concatPaths [ config.home directory ];
-                                        inherit persistentStoragePath;
-                                      }
-                                    else
-                                      directory // {
-                                        directory = concatPaths [ config.home directory.directory ];
-                                      });
                               };
                             };
                         }
@@ -295,7 +272,7 @@ in
                   };
 
                   files = mkOption {
-                    type = listOf (either str rootFile);
+                    type = listOf (coercedTo str (f: { file = f; }) rootFile);
                     default = [ ];
                     example = [
                       "/etc/machine-id"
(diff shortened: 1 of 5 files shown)


nixos.nix
```diff
@@ -436,26 +492,110 @@ in
...
+            # Parent directories of home folders. This is usually only
+            # /home, unless the user's home is in a non-standard
+            # location.
+            homeDirParents = mkParentDirs homeDirs;
+
+            # Parent directories of all explicitly listed directories.
+            parentDirs = mkParentDirs explicitDirs;
+
+            # All directories in the order they should be created.
+            allDirs = homeDirParents ++ homeDirs ++ parentDirs ++ explicitDirs;
```
tomeon, line 586: Maybe this list ought to be deduplicated by directory name?  Would potentially avoid setting/resetting directory ownership and permissions.

talyz: Well, we're already running `unique` for all the different directory types individually and they shouldn't overlap. We're checking for duplicate items already. The issues I can see are: conflicting parent dir specifications for files and recursive dir or file specifications, but both should be handled with assertions, not deduplicated. Maybe I'm missing something; do you have an example?

dschrempf:
This seems to be necessary. At the moment, Home Manager activation failes when using the impermanence system module and, e.g., `{ directory = ".local/share/mpd"; mode = "0700"; }`. The error is:
```
Apr 13 10:05:19 blaubaer hm-activate-dominik[1181]: mkdir: cannot create directory ‘/home/dominik/.local/share/applications’: Permission denied
Apr 13 10:05:19 blaubaer hm-activate-dominik[1182]: ln: failed to create symbolic link '/home/dominik/.local/share/applications/mimeapps.list': No such file or directory
```

m-bdf:
I've been using this branch for a bit and it works like a charm, avoiding the need to rely on systemd-tmpfiles to set correct parent permissions and ownership in most cases. Any chance of this PR being merged soon?

talyz:
Great! Yeah, I've been using this for a few months at this point without any issues, so I think it's safe to merge.

nixos.nix
tomeon: Hey; I know it's exceedingly, unbelievably, absurdly, laughably tardy, but:

> The issues I can see are: conflicting parent dir specifications for files and recursive dir or file specifications

Yes, this is precisely it.  Given a configuration like this:

```nix
environment.persistence = {
  "/abc" = {
    files = [
      { file = "/foo/bar"; parentDirectory.mode = "0755"; }
      { file = "/foo/lsgold"; parentDirectory.mode = "2700"; }
    ];
  };
};
```

We end up with a directory-creation script that looks like this:

```sh
#!/nix/store/5jw69mbaj5dg4l2bj58acg3gxywfszpj-bash-5.2p26/bin/bash
_status=0
trap "_status=1" ERR
/nix/store/ylachk42x2aa8x2hg4lx9gh70ks41igb-impermanence-create-directories '' '/abc' 'root' 'root' '0755' ''
/nix/store/ylachk42x2aa8x2hg4lx9gh70ks41igb-impermanence-create-directories '/abc' '/foo' 'root' 'root' '0755' ''
/nix/store/ylachk42x2aa8x2hg4lx9gh70ks41igb-impermanence-create-directories '/abc' '/foo' 'root' 'root' '2700' ''

exit $_status

```

This results in running `chown --reference=...`, `chmod --reference=...`, and various Bash file tests more than once.  Additionally, the mode `2700` is ignored, as `/abc/foo` already exists at the time of the third execution of the directory-creation script.

> [B]oth should be handled with assertions, not deduplicated.

Seems reasonable; I suggested deduplication as that would preserve the existing outcomes while cutting out unneeded operations.

PR #110, opened 2022-11-17
mount-file.bash: be quieter when debugging is off

ckiee:
Tested quickly on NixOS only with [this](https://github.com/ckiee/nixfiles/blob/f4b4002bd53070c625f11b49e1ad23a078c55ee9/modules/imperm.nix) config. 

Seems to work fine I think; for some reason the service output only sometimes got in my deploy log.


talyz:
Thanks!

PR #113, opened 2023-01-14
Add proper Home Manager module output

mk3z:


hxr404 (approved):
+1

Perhaps this change should be documented in the readme tho
(i.e. replace the suggestion to include via path with the new flake output)

hxr404:
> _No description provided._

**Add flake output for home-manager module**

The current / default way of importing the impermanence home-manager module feels kinda "hacky".
Using a flake output directly (instead of importing a specific .nix file using a path) is imo much more cleaner.
Additionally, in regards to future nixos caching / lazy eval efforts, this might make things easier, although I'm no expert at this.
Anyhow, I don't think there are any downsides to this change.

Yes, I know `impermanence.nixosModules.home-manager.impermanence` currently works too, but that also feels "hacky" and is kinda unreadable.
<hr />

since this has been stale for almost a month, I added a description to hopefully speed up review ;)
@talyz 
Would be nice if this got merged, I'm kinda autistic about my config :D
Btw: appreciate your (and everyone else's) work on this module. Thanks!

flake.nix
```diff
@@ -3,5 +3,6 @@
     nixosModules.impermanence = import ./nixos.nix;
     nixosModules.home-manager.impermanence = import ./home-manager.nix;
     nixosModule = self.nixosModules.impermanence;
+    hmModule = self.nixosModules.home-manager.impermanence;
```
otavio, line 6: I'd use `homeManager` as module name.

otavio (changes requested):

flake.nix
hxr404: I think sticking to hmModule would be better to make it consistent with the (admittedly, subjective) rest of the nixos ecosystem.

For instance, home-manager itself officially calls the type of a home-manager module `hmModule`
https://github.com/nix-community/home-manager/blob/ec06f419af79207b33d797064dfb3fc9dbe1df4a/nixos/common.nix#L14

And I've seen other projects in nix-community follow the same naming scheme, for example see
https://github.com/nix-community/nix-doom-emacs/blob/7d7eb4c02eff4228c344d564edeaaf0458d6a742/flake.nix#L126

I don't know whether that naming is ideal, however I think it's okay enough to not change it for consistency's sake.

And if we'd really want to deviate from this informal convention, something like `homeManagerModule` would imo be better suited (because it's a home-manager module being exported, not home-manager itself)

otavio (approved):

flake.nix
mk3z: Would this be better?
```suggestion
    homeManagerModule = self.homeManagerModules.impermanence;
```
And then change `nixosModules.home-manager.impermanence = import ./home-manager.nix;` to be `homeManagerModules.impermanence = import ./home-manager.nix;`.

ambroisie: The idiomatic way to signal a default for through a singular top-level attribute name has been deprecated to instead use `<plural-attr-name>.default`.

So IMO we should use `homeManagerModules.{impermanence,default}`.

talyz (changes requested):
Thanks for working on this and sorry about the slow response. Could you please document this flake output in the README as well? Just adding something which is similar to the [section for the NixOS module](https://github.com/nix-community/impermanence#nixos) should suffice.

SebTM:
Can be closed when #181 got merged after rebase now :)

PR #114, opened 2023-02-02
nixos: Recommend persisting /var/lib/nixos

mweinelt:
The `/var/lib/nixos` directory contains the uid and gid map for entities without a static id. Not persisting them means your user and group ids could change between reboots, which is likely undesirable.

### DIFF
--- a/README.org
+++ b/README.org
@@ -39,6 +39,7 @@
           directories = [
             "/var/log"
             "/var/lib/bluetooth"
+            "/var/lib/nixos"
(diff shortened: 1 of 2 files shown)


talyz:
Thanks!

jackinloadup:
Does this only affect systems with `users.mutableUsers = true`?

mweinelt:
It's about `users.users` and `users.groups` entries without statically allocated uid/gid.

PR #126, opened 2023-04-23
nixos: Fix collision detection logic

alois31:
The logic for detecting whether a file or directory is specified multiple times was not updated to account for the recent file/filePath and directory/dirPath changes. This can lead to spurious failures like the following if the same home-relative path is persisted for different users:

       Failed assertions:
       - environment.persistence:
           The following directories were specified two or more
           times:
             .cache
             .config
             .local/share
             .local/state
             .gnupg
             .ssh

In addition, the assertion may falsely not trigger in the contrived situation where the same entity is persisted both in a per-user configuration (using the relative path) and the global configuration (using the absolute path).

Fix these situations by checking the absolute paths again using filePath and dirPath.

### DIFF
--- a/nixos.nix
+++ b/nixos.nix
@@ -686,10 +686,10 @@ in
             '';
         }
         {
-          assertion = duplicates (catAttrs "file" files) == [ ];
+          assertion = duplicates (catAttrs "filePath" files) == [ ];
           message =
             let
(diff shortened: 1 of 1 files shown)


m-bdf (approved):

talyz:
Thanks!

PR #134, opened 2023-06-24
nixos: Introduce `persistentStoragePath` option

### LINKED ISSUE #117: Allow generic environment.persistence attribute names, rather than the persistent path
Hi, first of all thank you for the project 🏆

I have the following use-case:
In a NixOS configuration, I defined some paths to persist with the NixOS impermanence module, like this
```nix
environment.persistence = {
  "/persist/backup_me" = {
    directories = [ "/foo" "/bar" ];
  };
  "/persist/dont_backup" = {
    directories = [ "/fizz" "/buzz" ];
  };
};
```

Now, in another part of the configuration, I'd like to backup the content of `/persist/backup_me`. To do so, I need to get the path of the directory to backup "/persist/backup_me", so I'd write a line like this `backup.baseDir = "/persist/backup_me";`, duplicating the string "/persist/backup_me".

Would it be possible to change the module so that I did not need to specify the persistent path as the `environment.persistence` attribute name, but could instead specify an arbitrary value as the name and use a nested attribute, say `environment.impermanence.<name>.persistentPath` to specify the persistent path?

Impermanence could default to using `<name>` as the persistent path if `persistentPath` is not specified (like in NixOS for the [fileSystems attributes](https://github.com/NixOS/nixpkgs/blob/1f397d4e54c96df83820e99e87756637dad3133e/nixos/modules/tasks/filesystems.nix#L77)), so this would not be a breaking change for existing users.

The idea I have in mind is that in the end people could specify a NixOS configuration like this
```nix
environment.persistence = {
  backup = {
    persistentPath = "/persist/backup_me";
    directories = [ "/foo" "/bar" ];
  };
  "/persist/dont_backup" = {
    directories = [ "/fizz" "/buzz" ];
  };
};
# Hypothetical backup tool in which I specify the directory to backup
backup.baseDir = config.environment.persistence.backup.persistentPath;
```

Do you think this would be something worth adding?

talyz:
This allows the path to persistent storage to be defined through an option in the submodule, not just by the submodule name, i.e. instead of

```nix
  files = [ ... ]
  directories = [ ... ]
};
```

you could specify

```nix
environment.persistence.main = {
  persistentStoragePath = "/persistent";
  files = [ ... ]
  directories = [ ... ]
};
```

which is good for readability and when you want to split the config between multiple files.

Fixes #117.

cc @civts @eyJhb

### DIFF
--- a/nixos.nix
+++ b/nixos.nix
@@ -51,13 +51,13 @@ let
   '';
 
   # Create fileSystems bind mount entry.
-  mkBindMountNameValuePair = { dirPath, persistentStoragePath, ... }: {
+  mkBindMountNameValuePair = { dirPath, persistentStoragePath, hideMount, ... }: {
     name = concatPaths [ "/" dirPath ];
     value = {
       device = concatPaths [ persistentStoragePath dirPath ];
       noCheck = true;
       options = [ "bind" ]
-        ++ optional cfg.${persistentStoragePath}.hideMounts "x-gvfs-hide";
+        ++ optional hideMount "x-gvfs-hide";
       depends = [ persistentStoragePath ];
     };
   };
@@ -89,7 +89,6 @@ in
           submodule (
             { name, config, ... }:
             let
-              persistentStoragePath = name;
               defaultPerms = {
                 mode = "0755";
                 user = "root";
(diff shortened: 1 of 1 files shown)


mibmo:
Works for me! :) ETA on merging?

eyJhb:
I haven't had the time to test it out, and won't have time until Monday most likely... So if it works for others, just merge :D 

Sorry!

eyJhb:
Works fine! I think you can merge this :) 
Thanks for the doing this!

mibmo:
I second that! Been using it for a month with zero problems :)

eyJhb:
Any possibility for adding this to the home-manager module?

eyJhb:
I've added a PR that should add this. However, unsure of the code quality.

PR #135, opened 2023-08-17
home-manager: add persistentStoragePath option

eyJhb:
Add `persistentStoragePath` to the home-manager module, to be in line with #134 .

Would allow to do things such as...

```nix
home.persistence.main = {
  persistentStoragePath = "/persistent";
  files = [ ... ]
  directories = [ ... ]
};
```

cc @talyz :)

EDIT: this is not tested yet. Trying to use it now.
EDIT2: I've been running this for months now, without any issues.
EDIT3: I do however have patched the following line, to always use symlink `getDirMethod = v: v.method or "bindfs";`

### DIFF
--- a/home-manager.nix
+++ b/home-manager.nix
@@ -357,7 +357,7 @@ in
 
         mkBindMountsForPath = persistentStorageName:
           concatMapStrings
-            (mkBindMount persistentStoragePath)
+            (mkBindMount persistentStorageName)
             (map getDirPath (filter isBindfs cfg.${persistentStorageName}.directories));
 
         mkUnmount = persistentStorageName: dir:
--- a/home-manager.nix
+++ b/home-manager.nix
@@ -4,7 +4,7 @@ with lib;
 let
   cfg = config.home.persistence;
 
-  persistentStoragePaths = attrNames cfg;
+  persistentStorageNames = attrNames cfg;
 
   getDirPath = v: if isString v then v else v.directory;
   getDirMethod = v: v.method or "bindfs";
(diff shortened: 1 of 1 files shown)


eyJhb:
I'm really unsure, if this is the way to go. I've tried not to edit too much in the module, but only the base things.

eyJhb:
@SrEstegosaurio keep in mind, I've mostly only tested it with symlink, as I don't like bindfs. Hopefully your PC does not explode.

SrEstegosaurio:
Sorry for my late response, I've been busy. :/

I've decided to test it out following the example and got an error.

##### The nix expression:
```nix
{pkgs, lib, config, ...}:
{
	# Element matrix client configuration file.
	home = {
		# There is an element-desktop-wayland which currently does not spawn a window in wayland.
		# Maybe more testing?
		packages = [pkgs.element-desktop];

		persistence.main = lib.mkIf (builtins.elem pkgs.element-desktop config.home.packages) {
			persistentStoragePath = "/nix/persistent${config.home.username}"; 
			directories = [
				"${config.xdg.configHome}/Element"
			];
		};
	};
}
```

##### The error:
```sh
       error: The option `home-manager.users.srestegosaurio.home.persistence.main.persistentStoragePath' does not exist. Definition values:
       - In `/nix/store/9q12xmgzj1jifzy6gbl824ainfp649x3-source/modules/coms/element': "/nix/persistentsrestegosaurio"
       
```

eyJhb:
How are you importing the patch?
It seems weird that you would get that error. 

I've been running it since I made this pr, and haven't gotten that error. I can try to post my config once I get home.

But it just smells very much like the patch isn't imported/used.

eyJhb:
@talyz do you have time to look at this PR? :)

talyz:
Looks good! Thanks! :)

PR #144, opened 2023-10-13
Fix build with `documentation.nixos.includeAllModules = true;`

Ma27:
This includes the options of all modules used in the evaluation, not just the ones from `<nixpkgs/nixos>` in the local manual.

Right now this breaks with

    error: attribute '"‹name›"' missing

because the submodule `environment.persistence` doesn't have actual declarations when building the manual, but a dummy only to evaluate the sub-options (including their defaults which is the cause of the error) and generate documentation from that.

Using `defaultText` prevents that because there's no need anymore to evaluate the `default` values of the options.

cc @lovesegfault @talyz

### DIFF
--- a/nixos.nix
+++ b/nixos.nix
@@ -99,6 +99,7 @@ in
                   persistentStoragePath = mkOption {
                     type = path;
                     default = cfg.${name}.persistentStoragePath;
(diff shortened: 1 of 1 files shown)


lovesegfault:
Thanks!

PR #146, opened 2023-10-21
Allow for symlinking files

MagicRB:
I have been running with this for a while now, both the bind mount and symlink code path. It works perfectly

### DIFF
--- a/mount-file.bash
+++ b/mount-file.bash
@@ -10,7 +10,7 @@ shopt -s inherit_errexit  # Inherit the errexit option status in subshells.
 trap 'echo Error when executing ${BASH_COMMAND} at line ${LINENO}! >&2' ERR
 
 # Get inputs from command line arguments
-if [[ "$#" != 3 ]]; then
+if [[ "$#" != 4 ]]; then
     echo "Error: 'mount-file.bash' requires *three* args." >&2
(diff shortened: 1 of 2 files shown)


Ma27:
LGTM. Using that now (merged into master at https://github.com/Ma27/impermanence/tree/tracking) to properly persist my gdb history file.

cc @talyz

talyz:
Formatting fails for this and conflicts need to be resolved. I also think using the file mounting mechanism for this makes it a bit confusing - it probably deserves its own file and logic. Alternatively, files and functions should be renamed to reflect the new purpose.

mount-file.bash
```diff
@@ -10,14 +10,15 @@ shopt -s inherit_errexit  # Inherit the errexit option status in subshells.
 trap 'echo Error when executing ${BASH_COMMAND} at line ${LINENO}! >&2' ERR
 
 # Get inputs from command line arguments
-if [[ "$#" != 3 ]]; then
+if [[ "$#" != 4 ]]; then
     echo "Error: 'mount-file.bash' requires *three* args." >&2
```
descention, line 14:
```suggestion
    echo "Error: 'mount-file.bash' requires *four* args." >&2
```

PR #150, opened 2023-11-16
sanitizeName: don't clobber dots in names

### LINKED ISSUE #132: home-manager: some paths lead to systemd unit file name collisions
For example, the home-manager module does not create a service for `~/logseq` when `~/.logseq` is already configured. This seems to be because `~/logseq` and `~/.logseq` lead to the same systemd unit file name, with the dot removed.

anna328p:
Fixes #132.

After rebuilding with this commit, a reboot (or maybe a `systemctl --user daemon-reexec`?) is required to successfully load and start the new units.

### DIFF
--- a/lib.nix
+++ b/lib.nix
@@ -56,9 +56,7 @@ let
       parents;
 
   sanitizeName = name:
-    replaceStrings
-      [ "." ] [ "" ]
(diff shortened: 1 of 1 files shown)


talyz:
Fixed by #155.

PR #151, opened 2023-11-19
README: Expand with more details on system setup and a better intro

talyz:
Expand the readme with details on how to set up the ephemeral root and improve the intro a bit.


tcurdt:
exactly what I was searching for.

Issue #139, opened 2023-09-04
[SECURITY] Change default permissions for created user directories to a secure value


oddlama:
Many users assume that if the home directory itself uses strict permissions such as `700` or `750`, the contents will also not be accessible to other users. This is usually the case on classical setups, but when using impermanence, this assumption no longer holds since the source directories will be created with very loose permissions:

https://github.com/nix-community/impermanence/blob/e3a7acd113903269a1b5c8b527e84ce7ee859851/nixos.nix#L252-L256

A simple github search for `lang:nix persistence users .gnupg` reveals that there already are plenty of affected setups (used wrongly in about every second) that publicly expose access to user's private keys or session data to the system. Therefore, I believe this is a dangerous default and should ideally be changed to `700` or at least `750`.

oddlama:
Looking over this again I see that impermanence should have created the home directory in the persistent folders with 700 already: 
https://github.com/nix-community/impermanence/blob/e3a7acd113903269a1b5c8b527e84ce7ee859851/nixos.nix#L572-L581

Yet, for some reason mine ended up with 755. I'll have to investigate this more to see what happened.

TheRealGramdalf:
> Looking over this again I see that impermanence should have created the home directory in the persistent folders with 700 already:

I think this comes from the following lines:

https://github.com/nix-community/impermanence/blob/23c1f06316b67cb5dabdfe2973da3785cfe9c34a/nixos.nix#L569-L573

This is resulting in some issues where if you persist e.g. `~/.local/state/nix`, but not `~/.local` or ~/.local/state`, parent directories that exist in persistent storage but not the target (i.e. local and state) will be created with owner `root` and permissions `755` upon reboot.
Interestingly, simply doing a `rebuild` causes the ownership of created directories to be correct, so it's only upon reboot that permissions get set to root.

talyz:
`defaultPermissions` apply to the creation of parent directories, but shouldn't affect the contents of the directories themselves, so I don't think that has anything to do with this at all. Could be a script failure, but it's hard to say with so little info available. Do you have an example config that gives this result?

PR #163, opened 2024-02-04
Add capability to let impemanence create the `persistentStoragePath`

portiatrivisonno:
Add option `createPersistentStoragePath` (bool) which, if set, creates the directory `persistentStoragePath` during activation phase. The default is set to `true` as I think this is what most people expect and also avoids some permissions mismatches (see below). If you do not agree for whatever reason, feel free to drop the commit enabling it by default.

The permission mismatch might also be related to #139.

## More Background

Beginning of the story is, that I bootstrapped a new system with this config:
```nix
  environment.persistence."home" = {
    persistentStoragePath = "/persist/user/this/dir/does/not/yet/exist";
    users."me" = {
      directories = [
        "cooldir"
      ];
    };
  };
```

After rebooting, cooldir was created by owned by root. It took me a while but the problem is/was that impermanence failed to create `/home/me/cooldir` inside the `persistentStoragePath` location (because that directory does not exist).  However, a bind entry was created in `/etc/fstab` which triggered systemd-fstab-generator to generated that mount and in the process also create the missing directories but with root permissions.

With this PR the config could look like
```nix
  environment.persistence."home" = {
    persistentStoragePath = "/persist/user/this/dir/does/not/yet/exist";
    createPersistentStoragePath = true;
    users."me" = {
      directories = [
        "cooldir"
      ];
    };
  };
```

### DIFF
--- a/nixos.nix
+++ b/nixos.nix
@@ -248,6 +248,18 @@ in
                     description = ''
                       The path to persistent storage where the real
                       files and directories should be stored.
+
+                      This directory will be created if createPersistentStoragePath is set.
+                    '';
+                  };
+
+                  createPersistentStoragePath = mkOption {
+                    type = bool;
+                    default = false;
+                    description = ''
+                      If set and persistentStoragePath does not exist then the directory
+                      persistentStoragePath will be created.  The owner of the directory
+                      will be root.
                     '';
                   };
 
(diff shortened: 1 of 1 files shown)


talyz:
Fixed by #193.

PR #168, opened 2024-02-16
Fix issue with directories not being found

kinzoku-dev:
An error was happening to me on a new NixOS installation when booting for the first time. It said something about a persistent directory not being found, and I was able to rectify the error with this simple fix.


talyz:
Fixed by #193.

PR #172, opened 2024-02-24
nixos: Create persisted dirs marked as needed for boot before mount

### LINKED ISSUE #121: Mounting paths needed for boot can fail
Booting a system with `environment.persistence."/persistent".directories = [ "/var/lib/nixos" ]` (from the README) fails if /persistent/var/lib/nixos does not exist yet:

```
waiting for device /mnt-root/persistent/var/lib/nixos to appear.......................
Timed out waiting for device /mnt-root/persistent/var/lib/nixos, trying to mount anyway.
mounting /mnt-root/persistent/var/lib/nixos on /var/lib/nixos...
mount: mounting /mnt-root/persistent/var/lib/nixos on /mnt-root/var/lib/nixos failed: No such file or directory

An error occurred in stage 1 of the boot process, which must mount the
root filesystem on `/mnt-root` and then start stage 2. Press one
of the following keys:

  r) to reboot immediately
  *) to ignore the error and continue
```

The init script of boot stage 1 tries to mount non-existent /persistent/var/lib/nixos on /var/lib/nixos, because this path is [needed for boot](https://github.com/NixOS/nixpkgs/blob/5fd439e1ce737fdfb22159faf38fe0be504674e9/nixos/lib/utils.nix#L13-L14). However, the [activation script](https://github.com/nix-community/impermanence/blob/master/create-directories.bash) responsible for creating missing directories is only run after stage 1, so mounting fails.

This issue should occur for any path needed for boot, whose persistent storage does not exist at boot.

talyz:
Create directories which are persisted, but also happen to be marked as needed for boot, at early boot. This happens for a few paths which NixOS always marks as such if there's a `fileSystems`-entry for them, for example `/var/lib/nixos`.

Fixes #121.

### DIFF
--- a/nixos.nix
+++ b/nixos.nix
@@ -6,12 +6,15 @@ let
     attrValues
     zipAttrsWith
     flatten
+    mkAfter
+    mkIf
     mkOption
     mkDefault
     mapAttrsToList
     types
     foldl'
     unique
+    concatMap
     concatMapStrings
     listToAttrs
     escapeShellArg
@@ -33,6 +36,7 @@ let
 
   inherit (utils)
     escapeSystemdPath
+    fsNeededForBoot
     ;
 
   inherit (pkgs.callPackage ./lib.nix { })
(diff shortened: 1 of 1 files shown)


segator:
Hey interested on this due I trying to use nixos-anywhere + impermanence, can you provide an example of how to configure  a flake with this persisted paths so I can test, thanks!

talyz:
Hi! You can use the flake ref `github:nix-community/impermanence/create-needed-for-boot` to try this out.

segator:
is just to enable this in flake? I tought reading your change we needed to enable something else. nevertheless if this is the case it does not work I made it work by creating the folders mkdir in disko device post hook

talyz:
Yes, it should just work automatically with no extra config. I don't know how it would interact with nixos-anywhere or disko, though. Do you have a config / flake you can share? I need some more context to investigate why it doesn't work for you.

segator:
Hey sorry I completely out of time this weeks, as soon as I have I will prepare a simple flake you can fully test in a VM with nixos-anywhere. but defintelly does not work :(

visualphoenix:
flying in from #121 to say this did not help fix my provisioning issue with flake + nixos-anywhere + disko + impermanence not playing nicely. commenting impermanence out unblocks provisioning.

talyz:
The nixos-anywhere + disko use case looks unrelated to what this PR is attempting to solve and should probably have its own issue. Regardless, I would also need a flake or config that reproduces the issue to investigate it further.

Defelo:
Hi, thanks for the fix! I have successfully tested this PR both with and without systemd in initrd. Is there anything else that needs to be done before this can be merged?

Defelo:
No, probably because this branch was not up to date when it was merged, and mkIf has already been added on master.

PR #182, opened 2024-05-08
hm: option for default linking method

Misterio77:
Hi all!

Here's a (slightly late) follow up to https://github.com/nix-community/impermanence/pull/99.

This adds an option that allows users to change the default linking method.

Bindfs has been bothering me with the performance hits, and writing `method = symlink` for every entry is cumbersome. The default still is bindfs, so no breaking changes.

To make this implementation cleaner, I also refactored the directory type to use coercedTo (in its own commit, to be easier to review).

### DIFF
--- a/home-manager.nix
+++ b/home-manager.nix
@@ -6,10 +6,9 @@ let
 
   persistentStorageNames = (filter (path: cfg.${path}.enable) (attrNames cfg));
 
-  getDirPath = v: if isString v then v else v.directory;
-  getDirMethod = v: v.method or "bindfs";
-  isBindfs = v: (getDirMethod v) == "bindfs";
-  isSymlink = v: (getDirMethod v) == "symlink";
+  getDirPath = v: v.directory;
+  isBindfs = v: v.method == "bindfs";
+  isSymlink = v: v.method == "symlink";
 
   inherit (pkgs.callPackage ./lib.nix { })
     splitPath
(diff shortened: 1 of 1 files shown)


Misterio77:
Can confirm there are no diffs in my home path when running this branch versus the main branch

maxbfuer:
I've been using this PR for about a week now and everything is working great on my end. Thank you both for your work on this PR and project :)

home-manager.nix
```diff
@@ -56,26 +53,45 @@ in
...
+
+                  - bindfs is very transparent, and thus used as a safe
+                  default. It has, however, a significant performance impact in
+                  IO-heavy situations.
+
+                  - symlinks have great performance but may be treated
+                  specially by some programs that may e.g. generate
+                  errors/warnings, or replace them.
+
+                  This can be overrided on a per entry basis.
```
talyz, line 70:
```suggestion
                  This can be overridden on a per entry basis.
```

home-manager.nix
```diff
@@ -6,10 +6,7 @@ let
 
   persistentStorageNames = attrNames cfg;
 
-  getDirPath = v: if isString v then v else v.directory;
-  getDirMethod = v: v.method or "bindfs";
-  isBindfs = v: (getDirMethod v) == "bindfs";
-  isSymlink = v: (getDirMethod v) == "symlink";
+  getDirPath = v: v.directory;
```
talyz, line 9: Since you're removing all the other accessors, maybe this one should go, too?

talyz (changes requested):
Thanks a lot for your work on this! Looks great! Just some minor comments.

fbuchlak:
Hi @talyz @Misterio77 Don't want to necrobump, but is there something except minor comments blocking this PR? This feature would be handy.

Misterio77:
Hey, sorry for the delay! Got quite busy last semester. Will see if I can get this going.

home-manager.nix
Misterio77: Thanks!

home-manager.nix
Misterio77: Sounds good!

home-manager.nix
Misterio77: Addressed: https://github.com/nix-community/impermanence/compare/68ac02aa0266326820fa540216384447f38a37fe..8825ba211a9b079f9b0c45d74645cf81b2c82b4a

home-manager.nix
Misterio77: Addressed: https://github.com/nix-community/impermanence/compare/68ac02aa0266326820fa540216384447f38a37fe..8825ba211a9b079f9b0c45d74645cf81b2c82b4a

Misterio77:
Should be good to go!

CC @talyz @fbuchlak

Misterio77:
Oops sorry, hold on a bit. Got some eval errors after rebasing.

talyz:
Thanks!

eyJhb:
@Misterio77 thanks for this PR, I _REALLY_ appreciate it. I've just been patching impermanence for a long while, instead of doing it properly. Thanks for the effort! :)

PR #183, opened 2024-05-08
hm: remove occurences of 'with'

Misterio77:
Just a personal annoyance I had while working on these again haha. Context: https://nix.dev/guides/best-practices#with-scopes

### DIFF
--- a/home-manager.nix
+++ b/home-manager.nix
@@ -1,12 +1,23 @@
 { pkgs, config, lib, ... }:
 
-with lib;
 let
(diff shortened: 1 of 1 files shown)


talyz:
I agree, it would be nice to get rid of these `with`s and I think most of this looks good, but rather than prefixing with `lib.` for the types, I prefer the solution employed in `nixos.nix`: https://github.com/nix-community/impermanence/blob/master/nixos.nix#L73-L84

PR #188, opened 2024-06-05
nixos: Add assertion for persisting UIDs/GIDs

### LINKED ISSUE #178: The importance of persisting `/var/lib/nixos` is not documented
The examples do show `/var/lib/nixos` being persisted, but it's never mentioned that doing so is vital to the correct functioning of the UID/GID allocation mechanism of NixOS.

If `/var/lib/nixos` is not persisted, UIDs/GIDs are allocated sequentially, and so can shift around across boots if new ones are added to the configuration. It can happen that after a reboot, a file is now randomly owned by a different user.

The documentation should warn about this, and as a longer term solution I would suggest adding an assertion: Only allow `/var/lib/nixos` to not be persisted if all users/groups have explicitly declared ids.

kuruczgy:
Fixes #178

### DIFF
--- a/nixos.nix
+++ b/nixos.nix
@@ -687,6 +687,9 @@ in
         homeDirOffenders =
           filterAttrs
             (n: v: (v.home != config.users.users.${n}.home));
+        usersWithoutUid = attrNames (filterAttrs (n: u: u.uid == null) config.users.users);
+        groupsWithoutGid = attrNames (filterAttrs (n: g: g.gid == null) config.users.groups);
+        varLibNixosPersisted = elem "/var/lib/nixos" (catAttrs "dirPath" directories);
       in
       [
         {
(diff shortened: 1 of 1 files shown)


kuruczgy:
Gentle reminder @talyz. Do you in general like the idea of this PR? Do you think we could find someone to review this?

The corresponding issue I opened (#178) did get attention from a couple other people as well, so it's clear that I was not the only person who ran into this issue, and this check would be useful to have.

talyz:
Looks really useful! Thanks! :) And sorry for the slow response!

surfaceflinger:
It should be possible for this to be skipped by some option

https://github.com/nix-community/impermanence/commit/9de98e038ae91e15ea725700386044309b340299#commitcomment-145532610
https://github.com/nix-community/impermanence/commit/9de98e038ae91e15ea725700386044309b340299#commitcomment-145533686

nakoo:
I feel this merge needs to be more aware. At least it needs to be documented.

NickCao:
And we should check all possible parent directories, `/`, `/var` and `/var/lib`

merrkry:
When parent directories like `/var` or `/var/lib` are persisted, impermanence will give false alarm and prevent rebuilding.

accelbread:
There should be a way to turn this off.
On my system, I have my own users's uid and gid set, but for the rest of the users/groups, I have no dependency on what uid/gid they have since I dont persist any files from those users.

kuruczgy:
Oops, yeah, didn't think of that.

At least we got some quick feedback on this now (haha), but @talyz we should probably revert this as it's clearly breaking people's setups.

I might have time in a few days to work on a fixup. From the comments above I gather that:
1. We should check all parent dirs of `/var/lib/nixos`, this is obvious, sorry I missed this.
2. There should still be a way to turn it off.

If anyone has additional ideas for how a future attempt at this could be improved, please leave a comment.

talyz:
Yep, I agree with all of this. I've pushed a few changes to solve this now:
- changed the assertion to a warning, since it's not critical to most people
- include the parents directories in the check
- added an option to toggle warnings on and off (`enableWarnings`)

PR #189, opened 2024-06-09
nixos: Disable module if no persistent storage paths are defined or all are disabled

talyz:
Previously, we would always make some meaningless additions to `activationScripts` and such.

Also, remove some unused lib function imports from the module.

cc @0xf09f95b4

### DIFF
--- a/nixos.nix
+++ b/nixos.nix
@@ -8,6 +8,7 @@ let
     flatten
     mkOption
     mkDefault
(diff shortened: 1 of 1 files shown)


jian-lin:
Related: https://github.com/nix-community/impermanence/pull/190 tries to avoid making side-effects to `fileSystems` if `directories == []`

PR #190, opened 2024-06-16
nixos: avoid side effects to fileSystems if possible

jian-lin:
This is a good idea in itself.

Additionally, it avoids issues if users want to use mkDefault in their fileSystems config and have no persistent directories.

### DIFF
--- a/nixos.nix
+++ b/nixos.nix
@@ -506,9 +506,9 @@ in
       in
       foldl' recursiveUpdate { } (map mkPersistFileService files);
 
-    fileSystems = bindMounts;
+    fileSystems = mkIf (directories != [ ]) bindMounts;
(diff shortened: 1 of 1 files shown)


jian-lin:
@talyz Hi, I have fixed the formatting.  Please have look.

nixos.nix
```diff
@@ -506,9 +506,9 @@ in
       in
       foldl' recursiveUpdate { } (map mkPersistFileService files);
 
-    fileSystems = bindMounts;
+    fileSystems = mkIf (directories != [ ]) bindMounts;
```
jian-lin, line 509: Can also be written as
```suggestion
    fileSystems = mkIf (bindMounts != { }) bindMounts;
```

PR #193, opened 2024-06-23
nixos: Create persistent storage directories and their parents

### LINKED ISSUE #154: Fails with mkdir: cannot create directory: No such file or directory
I use impermanence with the following config:

```
  # Impermanence
  environment.persistence."/persist/system" = {
    directories = [
      "/etc/nixos" # bind mounted from /persist/system/etc/nixos to /etc/nixos
      "/etc/NetworkManager"
      "/etc/secureboot"
      "/var/log"
      "/var/lib"
    ];
    files = [
      "/etc/machine-id"
      "/etc/ssh/ssh_host_ed25519_key"
      "/etc/ssh/ssh_host_ed25519_key.pub"
      "/etc/ssh/ssh_host_rsa_key"
      "/etc/ssh/ssh_host_rsa_key.pub"
    ];
  };
```

When installing on a new system, I create a partition, mount it on `/persist`, and run nixos-install. However. this fails because when impermanence tries to create the persistence directories, it fails with `No such file or directory` errors.

I have fixed it [like this](https://github.com/nix-community/impermanence/compare/master...tmarkov:impermanence:master) - I'm nit sure though if this is a problem with impermanence or if I'm doing something wrong.

talyz:
If they're not filesystem roots, they won't exist when deploying to a new system and would otherwise have to be manually created.

Fix #154

cc @tmarkov

### DIFF
--- a/create-directories.bash
+++ b/create-directories.bash
@@ -46,14 +46,16 @@ fi
 # check that the source exists and warn the user if it doesn't, then
 # create them with the specified permissions
 realSource="$(realpath -m "$sourceBase$target")"
-if [[ ! -d "$realSource" ]]; then
+if [[ ! -d $realSource ]]; then
     printf "Warning: Source directory '%s' does not exist; it will be created for you with the following permissions: owner: '%s:%s', mode: '%s'.\n" "$realSource" "$user" "$group" "$mode"
     mkdir --mode="$mode" "$realSource"
     chown "$user:$group" "$realSource"
 fi
 
-[[ -d "$target" ]] || mkdir "$target"
+if [[ $sourceBase ]]; then
+    [[ -d $target ]] || mkdir "$target"
 
-# synchronize perms between source and target
-chown --reference="$realSource" "$target"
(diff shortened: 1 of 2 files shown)


tmarkov:
I'll try to try this tomorrow. Out of curiosity, why isn't simply passing `-p` to `mkdir` good enough?

talyz:
It opens up for directories being created with the wrong permissions and ownership if they're created in the wrong order. You would always have to consider the possible side effects of it, basically. Explicitly creating all directories in the hierarchy gives much better control and makes potential issues easier to pinpoint.

visualphoenix:
any reason why this removes quoting in the bash part?

tmarkov:
Sorry for the delay, had trouble setting up a test vm properly.

But everything seems to work fine (at least on my config).

talyz:
> any reason why this removes quoting in the bash part?

Only that quoting is unnecessary inside double square brackets, as string splitting is not performed in that context.

Issue #198, opened 2024-08-05
bindMount logic flaw: There is already an active mount at or below 'foler'!


PowderedToastMan:
Under my current configuration, the Impermanence module refuses to bindMount folders under my home directory because it mistakenly believes there is already and active mount.

I maintain a `/persist` folder containing all persistent data.  Under this folder, I mount a few zfs filesystems.
```
fileSystems = {
  "/persist/home/dave/Downloads" = { device = "tank/dave/Downloads"; fsType = "zfs"; };
  "/persist/home/dave/tmp" = { device = "tank/dave/tmp"; fsType = "zfs"; };
  "/persist/home/dave/Documents" = { device = "spool/dave/Documents"; fsType = "zfs"; };
  #...
};

persistence."/persist/home/${config.user}" = lib.mkIf config.impermanence {
  allowOther = true;
  directories = [
    ".cache"
    "Documents"
    "Downloads"
    "tmp"
  ];
  files = [
    ".local/share/zsh/history"
  ];
};
```

The folder fails to mount when activating home-manager, only creating an empty folder.  Upon closer inspection, it would appear that the logic that determines whether the destination is already mounted is mistakenly tripping on this setup.
```
$ /nix/store/ik7kg6fm1isyvi53xmbkkaxi466ai0yw-hm-setup-env /nix/store/g71f2hkfylxdwsqj45grx5455wfdqf1i-home-manager-generation
Starting Home Manager activation
Activating checkFilesChanged
Activating checkLinkTargets
Existing file '/home/dave/.local/share/zsh/history' is in the way of '/nix/store/gs0nfb9m210qh4rvmp1gdk9ybf51j332-home-manager-files/.local/share/zsh/history', will be skipped since they are the same
Activating unmountPersistentStoragePaths
Activating createAndMountPersistentStoragePaths
Activating createTargetFileDirectories
Activating writeBoundary
Activating createGpgHomedir
Activating installPackages
Activating linkGeneration
Cleaning up orphan links from /home/dave
No change so reusing latest profile generation 1
Creating home file links in /home/dave
Activating onFilesChange
Activating runUnmountPersistentStoragePaths
Activating reloadSystemd
The user systemd session is degraded:
  UNIT                                            LOAD   ACTIVE SUB    DESCRIPTION
● bindMount--persist-home-dave-Documents-.service loaded failed failed Bind mount '/persist/home/dave/Documents' at '/home/dave/Documents'
[...]

PR #205, opened 2024-08-27
Add warning regarding home-manager module

bphenriques:
Impermanence within home-manager will not work if  `~/.config/systemd` is added. 

See https://github.com/nix-community/impermanence/issues/198#issuecomment-2312122729

Ideally this should be an assertion instead to avoid building broken builds, would you like me to update?


bphenriques:
Not actively using the module and this should be an assertion

PR #181, opened 2024-04-27
Add proper Home Manager module output (Updated #113)

SebTM:
Hey, I took care of the requested change in #113 but don't had permission to update the MR so why I created this 🙏🏻 Let me know if you would like something different - rebased on latest master.

PR #211, opened 2024-09-09
homeManagerModules attripute naming outside nixosModules

siryoussef:
adding flake output homeManagerModules.impermanence just not to cause confusion if used with standalone home-manager flake setup (as one usually does not use nixosModules attribute in defining a homeManager module in such a case even though it would work here !, although keeping the old flake output ~nixosModules.home-manager.impermanence is a good idea for backward compatibility.

### DIFF
--- a/README.org
+++ b/README.org
@@ -349,6 +349,29 @@
       }
     #+end_src
 
+    or use the provided ~homeManagerModules.impermanence~ flake output:
+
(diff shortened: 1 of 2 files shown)


SebTM:
Duplicate of https://github.com/nix-community/impermanence/pull/181

PR #217, opened 2024-09-24
nixosModule: only inherit mkIf once

herrwiese:


herrwiese:
Whoops.  Duplicate of #214

PR #220, opened 2024-09-28
nixos: Fix create-needed-for-boot-dirs dependencies for zfs

### LINKED ISSUE #216: ZFS not compatible with PR #172
This PR (https://github.com/nix-community/impermanence/pull/172) is not compatible with ZFS. It relies on the device defined in `fileSystems.<name>.device`, but for instance, if `fileSystems."/persist".device = "zpool/persist"`, the corresponding `zpool-persist.device` won't exist in ZFS.

To make it work with ZFS, the code needs to be modified.

```nix
boot.initrd.systemd.services.create-needed-for-boot-dirs = {
  after = lib.mkForce ["zfs-import-<pool>.service"];
  wants = lib.mkForce ["zfs-import-<pool>.service"];
  unitConfig.DefaultDependencies = "no";
};
```
I'm not sure why `unitConfig.DefaultDependencies = "no"` wasn't set in the first place though.

talyz:
Fix #216.

cc @nakoo

### DIFF
--- a/nixos.nix
+++ b/nixos.nix
@@ -719,8 +719,14 @@ in
             matchFileSystems = fs: attrValues (filterAttrs (_: v: v.mountPoint or null == fs) all);
           in
           concatMap matchFileSystems persistentStoragePaths;
-        devices =
-          map (d: "${(escapeSystemdPath d)}.device") (catAttrs "device" fileSystems);
+        deviceUnits = unique
+          (map
+            (fs:
+              if fs.fsType == "zfs" then
+                "zfs-import.target"
+              else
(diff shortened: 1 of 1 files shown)


matdibu:
is this applicable to both legacy and non-legacy zfs mounts?

nakoo:
@talyz The system booted successfully as expected. Thank you for quick fix.

@matdibu I'm using `mountpoint=legacy` for my setup and haven't encountered any issues with this PR. However, I'm unsure how it affects non-legacy mounts.

matdibu:
@nakoo just tested it after the merge on a system with non-legacy mounts, it works fine 👍🏻

PR #221, opened 2024-09-29
nixos: Fix create-needed-for-boot-dirs label-adressed dependencies

### LINKED ISSUE #218: `fileSystems.<name>.label` not compatible with PR #172
When specifying a file system by label using `fileSystems."/foo".label = "foo";` instead of using the `device` attribute, this now fails to evaluate as PR #172 expects the latter to be set.  This is similar to the ZFS problem described in #216.

talyz:
When a file system entry is specified by its label, not by its device, we have to generate the device path from the label. While this was already done in the script, it wasn't for the systemd unit dependencies.

Fixes #218.

cc @herrwiese @MatthewCash

### DIFF
--- a/nixos.nix
+++ b/nixos.nix
@@ -695,10 +695,11 @@ in
       let
         neededForBootFs = catAttrs "mountPoint" (filter fsNeededForBoot (attrValues config.fileSystems));
         neededForBootDirs = filter (dir: elem dir.dirPath neededForBootFs) directories;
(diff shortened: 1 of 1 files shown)


MatthewCash (approved):
Works for me

herrwiese (approved):

herrwiese:
Looks good here, too.

PR #223, opened 2024-10-04
fix impermanence in combination with sysusers/userborn

Mic92:
Both remove user creation from the action phase to a systemd service. This is hopefully also easier to debug, since we now can look at a systemd service instead for log output.


create-directories.bash
```diff
@@ -48,12 +48,10 @@ fi
...
-    chown "$user:$group" "$realSource"
+    mkdir "$realSource"
 fi
 
 [[ -d "$target" ]] || mkdir "$target"
 
-# synchronize perms between source and target
-chown --reference="$realSource" "$target"
-chmod --reference="$realSource" "$target"
+chown "$user:$group" "$realSource" "$target"
```
Mic92, line 56: See commit message for reasoning.

Mic92:
Can be tested with `nix run github:Mic92/userborn-with-impermanence#nixosConfigurations.myhost.config.system.build.vmWithDisko`

create-directories.bash
talyz: I'm a bit skeptical of this, since it would apply mode changes to already existing directories, whether a mode is supplied for the directory or not. I'm not really sure what's not working for you, though. Are the initial directories created with the incorrect permissions?

talyz (changes requested):
Looks like you're moving a lot of code from one scope to another, but it's intermixed with your actual changes. Please split out the move to its own commit, so that it's easier to review the changes.

create-directories.bash
talyz: Also, I guess this isn't critical for the main purpose of this PR, so maybe it could be moved to its own?

Issue #237, opened 2024-12-01
Boot dependencies broken when persistence directory is a bind mount


Majiir:
Consider this simplified example:

```nix
{
  fileSystems = {
    "/nixos" = {
      device = "/dev/sda1";
      fsType = "ext4";
      neededForBoot = true;
    };

    "/persist" = {
      device = "/nixos/persist";
      options = [ "bind" ];
    };
  };

  environment.persistence."/persist" = {
    directories = [
      "/var/lib/nixos"
    ];
  };
}
```

When using systemd initrd, impermanence creates a `create-needed-for-boot-dirs.service` unit with `Requires=nixos-persist.device` and `Wants=nixos-persist.device`. No such device exists, so boot hangs until the device unit times out.

The relevant logic is [here](https://github.com/nix-community/impermanence/blob/3ed3f0e/nixos.nix#L752-L759).

We should check whether the filesystem is a bind mount and create the unit name accordingly. NixOS has [this check](https://github.com/NixOS/nixpkgs/blob/8b125fe422e792de8211b305328500e12549ea2f/nixos/modules/tasks/filesystems.nix#L197) for whether a filesystem is a bind mount.

Luminger:
I'm experiencing the same hangs you describe but with a different setup. I'm using `microvm.nix` with `impermanence`, where `/persist`:

```nix
{
  microvm.shares = [
    {
      source = "/mnt/pool/vm/persist";
      mountPoint = "/persist";
      tag = "persist";
      proto = "virtiofs";
    }
  ];
}
```

The issue you described is the same, `persist.device` units are placed into `Requires=` and `Wants=`.

The fix you proposed doesn't work for me as the backing filesystem is `virtiofs` and there's no bind option present. It would require another check to see if the filesystem is `virtiofs`, but this doesn't scale for other filesystems (like `p9`) that also lack devices. I think a generalized solution would be required here.

frantisekhanzlikbl:
Another useful filesystem with no device units is `tmpfs` (plus `ramfs`, whatever one might need that for). A general solution really is needed.

I'm using the following patch to work around this temporarily:
```patch
diff --git a/nixos.nix b/nixos.nix
index cdad872..a633edc 100644
--- a/nixos.nix
+++ b/nixos.nix
@@ -750,12 +750,11 @@ in
               in
               concatMap matchFileSystems persistentStoragePaths;
             deviceUnits = unique
-              (map
+              (concatMap
                 (fs:
[...]

PR #238, opened 2024-12-01
nixos: depend on mount unit for bind mounts

Majiir:
Fixes https://github.com/nix-community/impermanence/issues/237.


talyz:
Thanks!

Issue #147, opened 2023-11-03
Certain Files & Directories Used by GNOME Don't Work


jacobranson:
Hello! My NixOS config is using impermanence with an impermanent home directory. This is working mostly well, with a few things that I seemingly cannot get to work, all related to GNOME. The following files are correctly persisted, but GNOME seems to overwrite them with normal files instead of the bindmounted ones regardless.

- `.config/gnome-initial-setup-done` - for removing the "Welcome to GNOME" popup on first boot
- `.local/share/recently-used.xbel` - for persisting recently used files in Nautilus
- `.config/monitors.xml` - for persisting monitor configurations in GNOME Display Settings

Additionally, Trash support in Nautilus seems broken when bindmounting the `.local/share/Trash` directory.

Is anyone else using an impermanent home with GNOME and has gotten these things working? Thanks!

Jo-Blade:
Same issues here.
- recently-used.xbel and monitors.xml don't work when bind-mount by the module. But in general, mounting single files cause a lot of issues because programs tried to make copies for example so I am not really surprised.
- .local/share/Trash does not work with nautilus at all when using a symbolic link: files does not appear in trash and I can't put new files in trash. When using bind-mount, I can see files inside ~/.local/share/Trash inside the nautilus bin, but I can't add new files in the bin from nautilus. I can't figure why but it is very frustating…
- .config/gnome-initial-setup-done works, but is actually **useless because gnome do not use this file to remove Welcome to GNOME popup but a dconf key** the good solution is adding this inside home-manager :

```
    dconf.settings = {

      # Désactiver la popup de bienvenue de Gnome
      "org/gnome/shell" = {
        welcome-dialog-last-shown-version = "999"; # prevent popup until gnome version 999 :)
      };

    };
```

maydayv7:
> Additionally, Trash support in Nautilus seems broken when bindmounting the `.local/share/Trash` directory.

The FreeDesktop Specification mentions that the `~/.local/share/Trash` directory mustn't be a symbolic link in order to be used as intended. To preserve it's functionality, I've been using [this](https://github.com/maydayv7/dotfiles/blob/32f2d8f41cbfdd2bbcfa036f3db11355fc3bcba5/modules/hardware/filesystem.nix#L206) workaround instead of persisting that directory using `impermanence`

MordragT:
[...]

PR #243, opened 2024-12-19
Add an `allowTrash` option set to `true` by default to the NixOS module

bluelinden:
I ran into this issue about twenty minutes ago and tried my hand at fixing it. This resolves _parts of_ #147.

Newer versions of glib's trash implementation support an `x-gvfs-trash` mount option on bind mounts, that overrides the "internal filesystem" error that some people in #147 have been facing. 

This error is basically glib refusing to put things into a bind-mounted trash folder. I have added an option to the NixOS module that fixes this error by default, and is opt-out, but I would be fine making it opt-in if that fits the goals of the project better. I've confirmed that this works on my machine, but if people could test it out on theirs that would be great!

Please let me know if I've missed anything - I'm still relatively new to open-source and I am very tired right now :)

Relevant merge requests:
https://gitlab.gnome.org/GNOME/glib/-/merge_requests/4155
https://gitlab.gnome.org/GNOME/gvfs/-/merge_requests/222

Relevant issues:
https://github.com/nix-community/impermanence/issues/147
https://gitlab.gnome.org/GNOME/glib/-/issues/1885

### DIFF
--- a/README.org
+++ b/README.org
@@ -202,6 +202,7 @@
         environment.persistence."/persistent" = {
           enable = true;  # NB: Defaults to true, not needed
           hideMounts = true;
+          allowTrash = true;
           directories = [
             "/var/log"
             "/var/lib/bluetooth"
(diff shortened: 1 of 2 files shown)


bluelinden:
Hm, it may be better if this is set to false by default, because it leaves little `.Trash-1000`s at every mountpoint.

Issue #250, opened 2025-01-02
FS corruption after hibernation


xzfc:
The BTRFS script provided in the [README.org] has some problems that could result in filesytem corruption or data loss.

1. **Filesystem corruption.**
   This script runs both during the normal boot and when resuming from hibernation.
   Mounting BTRFS right before resuming is equivalent to mounting the same filesystem twice at the same time (without proper safeguards).
   I had my filesystem corrupted and turned into an unmountable/unbootable state because of this.

   The solution: instead of `postDeviceCommands`, use `postResumeCommands`. See [`stage-1-init.sh` source] for the difference.

   Also, see "How to fix the FS corruption after hibernation" below.

2. **Wiping the wrong subvolumes.**
   The function `delete_subvolume_recursively` might delete all the subvolumes on a disk (not just a particular subdirectory of `/old_roots`).
   This can occur if `old_roots` and its children are plain directories rather than subvolumes, which can happen after restoring disk contents from a backup.

   The reason is that the `-o` option in `btrfs subvolume list` prints the subvolumes of the given parent subvolume, not the subvolumes of the given directory.
   To illustrate:

   ```console
   $ ls
   nix  old_roots  persistent  root

   $ sudo btrfs subvolume list -o .        # note no subvolumes in `old_roots`
   ID 259 gen 2575 top level 5 path root
   ID 260 gen 2334 top level 5 path nix
   ID 261 gen 2575 top level 5 path persistent

   $ sudo btrfs subvolume list -o ./root        # root is a subvolume, the output is expected
   ID 262 gen 1862 top level 259 path root/srv
   ID 263 gen 1865 top level 259 path root/var/lib/portables
   ID 264 gen 1865 top level 259 path root/var/lib/machines
   ID 265 gen 2511 top level 259 path root/tmp
   ID 266 gen 2511 top level 259 path root/var/tmp

   $ sudo btrfs subvolume list -o ./old_roots/2024-11-20_05:35:10      # (!) not a subvolume, the output lists unrelated subvolumes
   ID 259 gen 2575 top level 5 path root
   ID 260 gen 2334 top level 5 path nix
   ID 261 gen 2575 top level 5 path persistent
   ```

   In this case, `delete_subvolume_recursively /old_roots/2024-11-20_05:35:10` will happily delete `/root`, `/nix`, and `/persistent`.

   Possible solutions:
[...]

PR #269, opened 2025-04-25
docs(README): Add guardrails against accidentally GCing last snapshot

workflow:
This is a solution for problem 3. decribed here: https://github.com/nix-community/impermanence/issues/250#issue-2765986966

cc: @xzfc

### DIFF
--- a/README.org
+++ b/README.org
@@ -125,9 +125,16 @@
               btrfs subvolume delete "$1"
           }
 
-          for i in $(find /btrfs_tmp/old_roots/ -maxdepth 1 -mtime +30); do
-              delete_subvolume_recursively "$i"
-          done
+          ###
+          # GC
+          ###
+          latest_snapshot=$(find /btrfs_tmp/persist/old_roots/ -mindepth 1 -maxdepth 1 -type d | sort -r | head -n 1)
+          # Only delete old snapshots if there's at least one that will remain after deletion
+          if [ -n "$latest_snapshot" ]; then
+              for i in $(find /btrfs_tmp/persist/old_roots/ -mindepth 1 -maxdepth 1 -mtime +30 | grep -v -e "$latest_snapshot"); do
+                  delete_subvolume_recursively "$i"
+              done
+          fi
 
           btrfs subvolume create /btrfs_tmp/root
           umount /btrfs_tmp


README.org
```diff
@@ -125,9 +125,16 @@
               btrfs subvolume delete "$1"
           }
 
-          for i in $(find /btrfs_tmp/old_roots/ -maxdepth 1 -mtime +30); do
-              delete_subvolume_recursively "$i"
-          done
+          ###
+          # GC
+          ###
+          latest_snapshot=$(find /btrfs_tmp/persist/old_roots/ -mindepth 1 -maxdepth 1 -type d | sort -r | head -n 1)
```
maxhbr, line 131:
```suggestion
          latest_snapshot=$(find /btrfs_tmp/old_roots/ -mindepth 1 -maxdepth 1 -type d | sort -r | head -n 1)
```

README.org
```diff
@@ -125,9 +125,16 @@
...
+          # Only delete old snapshots if there's at least one that will remain after deletion
+          if [ -n "$latest_snapshot" ]; then
+              for i in $(find /btrfs_tmp/persist/old_roots/ -mindepth 1 -maxdepth 1 -mtime +30 | grep -v -e "$latest_snapshot"); do
```
maxhbr, line 134:
```suggestion
              for i in $(find /btrfs_tmp/old_roots/ -mindepth 1 -maxdepth 1 -mtime +30 | grep -v -e "$latest_snapshot"); do
```

fabiancholewinski1234:
A simpler solution would be:
```
for i in $(find /btrfs_tmp/old_roots/ -mindepth 1 -maxdepth 1 -mtime +30 | sort | head -n -1); do
    delete_subvolume_recursively "$i"
done
```

Note the minus before `1` in the argument of `head`.  From head(1):
```
       -n, --lines=[-]NUM
              print the first NUM lines instead of the first 10;
              with the leading '-', print all but the last NUM lines of each file
```

Change to `-5` to keep at least 5 backups.

Remark: Since we `sort` the findings, the naming scheme of the snapshot directories must be such they are sorted chronologically with the latest snapshot last.  This is the case for the default naming scheme `YYYY-MM-DD_HH:MM:SS` in the README.

PR #249, opened 2024-12-31
fix: missing binary in activation script

DaRacci:
Had an issue when moving some directories over to symlink after being bindfs mounts, encountered this error on my next home-manager activation.

This patch should ensure that `fusermount` is not missing by referencing a binary within the store.

Log:
```
warning: the following units failed: home-manager-racci.service
× home-manager-racci.service - Home Manager environment for racci
     Loaded: loaded (/etc/systemd/system/home-manager-racci.service; enabled; preset: ignored)
     Active: failed (Result: exit-code) since Tue 2024-12-31 18:34:36 AEDT; 260ms ago
 Invocation: 93b46f796df94e409da03faadbe01b38
    Process: 1562829 ExecStart=/nix/store/m91h9kmjw78swvldfb9px8qnv1nb23ax-hm-setup-env /nix/store/lshgw95gk6rkyaw5kq3rcypq3g9rrj72-home-manager-generation (code=exited, status=127)
   Main PID: 1562829 (code=exited, status=127)
         IP: 0B in, 0B out
         IO: 64K read, 0B written
   Mem peak: 6.4M
        CPU: 83ms

Dec 31 18:34:34 nixe hm-activate-racci[1562829]: Activating checkFilesChanged
Dec 31 18:34:34 nixe hm-activate-racci[1562829]: Activating cleanEmptyLinkTargets
Dec 31 18:34:34 nixe hm-activate-racci[1562895]: /nix/store/lshgw95gk6rkyaw5kq3rcypq3g9rrj72-home-manager-generation/activate: line 250: fusermount: command not found
Dec 31 18:34:35 nixe hm-activate-racci[1563086]: /nix/store/lshgw95gk6rkyaw5kq3rcypq3g9rrj72-home-manager-generation/activate: line 250: fusermount: command not found
Dec 31 18:34:36 nixe hm-activate-racci[1563088]: /nix/store/lshgw95gk6rkyaw5kq3rcypq3g9rrj72-home-manager-generation/activate: line 250: fusermount: command not found
Dec 31 18:34:36 nixe hm-activate-racci[1562829]: Couldn't perform regular unmount of /home/racci/.config/Nextcloud. Attempting lazy unmount.
Dec 31 18:34:36 nixe hm-activate-racci[1563089]: /nix/store/lshgw95gk6rkyaw5kq3rcypq3g9rrj72-home-manager-generation/activate: line 256: fusermount: command not found
[...]

PR #272, opened 2025-05-25
Home Manager module v2, systemd mounts, options refactoring, tests and more

### LINKED ISSUE #45: home-manager systemd user units aren't auto-started?
I'm on a new box setting up impermanence for the first time.

I'm having to move some dirs out of the way to get `home-manager-cole.service` to activate properly.

However, it did start, but the actual bind mount units are all dead/inactive. I've had to start them myself?

### LINKED ISSUE #130: home-manager bind mounts gone missing.
Lately whenever I reboot, some or all of my home-manager bind mounts fail to be created. This can always be fixed by running `sudo systemctl restart home-manager-$USER.server`, but I have no idea why it's happening.

### LINKED ISSUE #142: Feature disparity between the NixOS and h-m modueles
I've found that there's an abysmal difference between the feature set of the both modules provided by this flake.

Why is it that you cannot change the permission, the persistentStorage path, etc... while in the home-manager module?

The most frustrating part is that, by default, all directory paths  in the m-m mod. are relative to the the user home. Is there a reason for this?

It is a problem if you want to, for example, so things like this one:

```nix
home.impermanence."/nix/persistence" = {
  directories = [
    config.xdg.dataHome;
  ]:
};
```

Which can be very useful in a lot of situations.

Thanks for you're time and this project. :)

### LINKED ISSUE #177: Symlinks created via the home-manager module are broken
Adding the following configuration creates a broken symlink `persistent` in `/home/janrupf`:

```nix
home.persistence."/persistent/home/janrupf" = {
  directories = [
    { directory = "persistent"; method = "symlink"; }
  ];
}
```

Inspecting:
```
$ readlink persistent
/nix/store/vlidns7s3sp46bkw65nmr8wj7lbhvkvx-home-manager-files/persistent
$ readlink /nix/store/vlidns7s3sp46bkw65nmr8wj7lbhvkvx-home-manager-files/persistent
/nix/store/iwbbjrdnqrffd8bjv4yi5yvanhj6w8f0-persistent-home-janrupf-persistent
$ readlink /nix/store/iwbbjrdnqrffd8bjv4yi5yvanhj6w8f0-persistent-home-janrupf-persistent
/persistent/home/janrupf/persistent
```

So the symlink does point to the correct location afterall - the issue is, that location is never created.
The creation is supposed to happen here:
https://github.com/nix-community/impermanence/blob/a33ef102a02ce77d3e39c25197664b7a636f9c30/home-manager.nix#L451-L461

Inspecting the flake output reveals that the command added to `home.activation.createTargetFileDirectories` is `mkdir -p /persistent/home/janrupf/.` - notice the dot at the end. I think this is caused by the `dirOf` call on line 458.

### LINKED ISSUE #179: Cannot use `users.users` without infinite recursion
I'd like to configure persistent directories for all normal users.

I figured a `filterAttrs` and `mapAttrs` would do the trick, however it seems `environment.persistence.*.users` modifies `users.users` in some way resulting in infinite recursion.

```nix
environment.persistance."/persist".users =
  mapAttrs
  (name: {
    directories = {
      # Defaults for all normal users
    };
  })
  (filterAttrs (name: user: user.isNormalUser) config.users.users);
```

I guess this the same underlying issue that means `environment.persistence.*.users.home` can't default to `users.users.*.home`? Although I'm unsure why that doesn't also apply to `users.users.*.group`...

It'd be great if _either_ the recursion issue could be somehow fixed, or at least better documented.
Alternatively providing an "allNormalUsers" option (terrible name, I know) that is equivalent to doing this, but somehow bypassing recursion issues, would also resolve this.

### LINKED ISSUE #184: home-manager module cannot mount files to /persist. permission denied.
i am having a nightmare trying to get the home-manager impermanence module to work. keep getting permission denied error and it does not appear to be mounting the home files to my /persist/home location. i am trying below but it is not a good approach and still i get error on the files now 'failed to create symbolic link '/home/billy/.config/gh/config.yml': Permission denied'. what am i doing wrong? is there a better way to do this? do i get same error if i was to use zfs over btrfs?

```

  systemd.tmpfiles.rules = [
    "d /persist/system/var/lib/nixos 0755 root root -"
    "d /persist/system/etc/nixos 0755 root root -"
    "d /persist/system/var/log 0755 root root -"

"d /persist/home/.local/share/docker 0755 root root -"
"d /persist/home/.config/VSCodium 0755 root root -"
"d /persist/home/.vscode 0755 root root -"
"d /persist/home/.vscode-oss 0755 root root -"
"d /persist/home/.config/libvirt 0755 root root -"
"d /persist/home/.zsh_history 0755 root root -"
"d /persist/home/.config/rclone 0755 root root -"
"d /persist/home/.config/rclone-browser 0755 root root -"
"d /persist/home/.config/qutebrowser/bookmarks 0755 root root -"
"d /persist/home/.config/qutebrowser/greasemonkey 0755 root root -"
"d /persist/home/.local/share/qutebrowser 0755 root root -"
"d /persist/home/.password-store 0755 root root -"
"d /persist/home/.config/nvim 0755 root root -"
"d /persist/home/.local/share/nvim 0755 root root -"
"d /persist/home/.local/state/nvim 0755 root root -"
"d /persist/home/tutorials 0755 root root -"
"d /persist/home/.config/go 0755 root root -"
"d /persist/home/.config/dlv 0755 root root -"
"d /persist/home/.virtualenvs 0755 root root -"
"d /persist/home/.gradle 0755 root root -"
"d /persist/home/notes 0755 root root -"
"d /persist/home/.pki 0755 root root -"
"d /persist/home/.cache/google-chrome 0755 root root -"
"d /persist/home/.cache/fontconfig 0755 root root -"
"d /persist/home/.config/google-chrome 0755 root root -"
"d /persist/home/.config/git 0755 root root -"
"d /persist/home/.config/gh 0755 root root -"
"d /persist/home/.gnupg 0755 root root -"


    
  ];
  
```

example of what i am doing for home persistence. is this not following best practices?
```
  home.persistence."/persist/home" = {
    directories = [
      ".config/qutebrowser/bookmarks"
      ".config/qutebrowser/greasemonkey"
      ".local/share/qutebrowser"
    ];
  };
```

### LINKED ISSUE #198: bindMount logic flaw: There is already an active mount at or below 'foler'!

### LINKED ISSUE #207: home-manager: mounts are not ordered by depth
looks like the [`createAndMountPersistentStoragePaths`](https://github.com/nix-community/impermanence/blob/c7f5b394397398c023000cf843986ee2571a1fd7/home-manager.nix#L421-L427) activation script orders mounts by persistence's names instead of mountpoints, meaning a parent directory can be attempted to be mounted AFTER all children were already mounted.

### LINKED ISSUE #222: `create-needed-for-boot-dirs` service fails on encrypted bcachefs filesystem
The service introduced in #172 seems to assume that once the device for a filesystem is available, the filesystem will be able to be mounted. This is not true in some cases, and the one I'm hitting is with bcachefs. For encrypted bcachefs, a separate systemd service is used to unlock the device, after which it's able to be mounted. `create-needed-for-boot-dirs` is not waiting for this unlock to happen, so it fails on every boot.

### LINKED ISSUE #227: `create-needed-for-boot-dirs` fails to umount properly
## Problem
I am new to using impermanence and I observed the following error with the `create-needed-for-boot-dirs` service

```shell
Oct 22 18:50:04 azathoth systemd[1]: Starting create-needed-for-boot-dirs.service...
Oct 22 18:50:04 azathoth create-needed-for-boot-dirs-start[1340]: umount: /persist-tmp-mnt/persist: target is busy.
Oct 22 18:50:04 azathoth systemd[1]: create-needed-for-boot-dirs.service: Main process exited, code=exited, status=32/n/a
Oct 22 18:50:04 azathoth systemd[1]: create-needed-for-boot-dirs.service: Failed with result 'exit-code'.
Oct 22 18:50:04 azathoth systemd[1]: Failed to start create-needed-for-boot-dirs.service.
```

## Configuration

I am using disko and zfs with native encryption to setup my disks. Given that I am using native encryption I cannot use `postDeviceCommands` with impermanence because my drives must be first unencrypted to rollback. Instead of `postDeviceCommands` I am using `initrd.systemd`. My service is defined below.

```nix
    initrd.systemd = {
      enable = true;
      services.rollback = {
        description = "Rollback ZFS datasets to a pristine state";
        serviceConfig.Type = "oneshot";
        unitConfig.DefaultDependencies = "no";
        wantedBy = ["initrd.target"];
        after = ["zfs-import-zroot.service"];
        requires = ["zfs-import-zroot.service"];
        before = ["sysroot.mount"];
        path = with pkgs; [zfs];
        script = ''
          zfs rollback -r zroot/root@empty && echo "rollback complete"
        '';
      };
      services.create-needed-for-boot-dirs = {
        after = pkgs.lib.mkForce ["rollback.service"];
        wants = pkgs.lib.mkForce ["rollback.service"];
      };
    };
```

As you can see I also added after and wants to `create-needed-for-boot-dirs` in case the rollback service was conflicting with `create-needed-for-boot-dirs`'s ability to umount.

Here is my impermanence configuration:
```nix 
{
  environment.persistence."/persist" = {
    directories = [
      "/var/log"
      "/var/lib/libvirt"
      "/var/lib/nixos"
      "/var/lib/systemd/coredump"
    ];
  };
  users.groups.multimedia = {};
  environment.persistence."/persist/save" = {
    directories = [
      {
        directory = "/media";
        group = "multimedia";
        mode = "0770";
      }
    ];
    users.collin = {
      directories = [
        "brew"
        "projects"
        "work_projects"
        "Downloads"
        "Pictures"
        "Documents"
        "Videos"
        {
          directory = ".config/sops/age/";
          mode = "0700";
        }
        {
          directory = ".ssh";
          mode = "0700";
        }
        ".local/share/direnv"
      ];
    };
  };
}
```

It might also be relevant to show how my zfs datasets are defined:
```nix
        datasets = {
          persist = {
            type = "zfs_fs";
            options.mountpoint = "legacy";
            mountpoint = "/persist";
            options."com.sun:auto-snapshot" = "false";
            postCreateHook = "zfs snapshot zroot/persist@empty";
          };
          # dataset where all files that should both persist and need to be backed up are stored
          # the Impermanence module is what ensures files are correctly stored here
          persistSave = {
            type = "zfs_fs";
            options.mountpoint = "legacy";
            mountpoint = "/persist/save";
            options."com.sun:auto-snapshot" = "false";
            postCreateHook = "zfs snapshot zroot/persistSave@empty";
          };
          # Nix store etc. Needs to persist, but doesn't need backed up
          nix = {
            type = "zfs_fs";
            options.mountpoint = "legacy";
            mountpoint = "/nix";
            options = {
              atime = "off";
              canmount = "on";
              "com.sun:auto-snapshot" = "false";
            };
            postCreateHook = "zfs snapshot zroot/nix@empty";
          };
          root = {
            type = "zfs_fs";
            options.mountpoint = "legacy";
            options."com.sun:auto-snapshot" = "false";
            mountpoint = "/";
            postCreateHook = ''
              zfs snapshot zroot/root@empty
            '';
          };
        };
 ```
 
 My entire machine config can be found here https://github.com/collinarnett/brew/tree/main/hosts/azathoth

 ## Attempted fixes
 
 As I eluded to earlier, I tried to make sure that `create-needed-for-boot-dirs` doesn't conflict with the rollback service but I still have the same error.
  I also found [this commit](https://github.com/iynaix/dotfiles/commit/90c878a0438aec69481cc36414f0b392ac61b7cc) where the author explains
 > The generated script for the service unmounts /persist before /persist/cache and therefore errors out with a "target is busy" error. 
 
 This seems to imply that the order of unmounting matters but I'm not enough of an expert to determine a fix.

### LINKED ISSUE #232: Cannot make home mount if a seperate mount exists for the persistent path
There is an issue I believe with the code for the home fuse mount service, specifically here:

https://github.com/nix-community/impermanence/blob/0d09341beeaa2367bac5d718df1404bf2ce45e6f/home-manager.nix#L253

The issue is that if a user has an already existing mount that contains `mountPoint` as a substring, this service will fail with `There is already an active mount at or below ${mountPoint}!`

I believe this check was intended to ensure that the existing mount won't be overridden, but it fails in the case I describe above. I'm not sure what a fix would look like, but I think it might be to add a space to the beginning of the grep like so:
```sh
if ! mount | grep -F ' '${mountPoint}' ' && ! mount | grep -F ${mountPoint}/; then
```

The specific mount that the grep is finding in my case is a bind mount I've configured from `/persistent/home/z/.local/share/lutris` to`/mnt/data/home/z/LutrisShare`:
```
/dev/sda1 on /persistent/home/z/.local/share/lutris type ext4 (rw,relatime,x-systemd.requires-mounts-for=/persistent,x-systemd.requires-mounts-for=/mnt/data)
```
The nix config for the bind is:
```nix
  fileSystems."/persistent/home/z/.local/share/lutris" = {
    depends = [
      "/persistent"
      "/mnt/data"
    ];
    device = "/mnt/data/home/z/LutrisShare";
    fsType = "none";
    options = [ "bind" ];
  };
```

The reason I even need this bind in the first place is because `/mnt/data` is an entirely separate disk than what my `/persistent` is normally on. I wanted the actual data to be stored on this seperate disk rather than the disk that `/persistent` is on, and this is the only way I could think to do that.

### LINKED ISSUE #247: How to set parent dir's permissions for files in home-manager?
In the nixos module, there are userDefaultPerms and parentDirectory for files to set permissions. But I cannot find a one in the home-manager mode. Is there a way to set it?

### LINKED ISSUE #248: Significant Performance Impact
I've been running Impermanence for a long time now but have realised that the performance I'm getting really doesn't match what my hardware is capable of.

For reference my system flake can be found [here](https://github.com/DaRacci/nix-config) and I'm on the nixe system with the relevant files to these mounts being located in these files [/var/log](https://github.com/DaRacci/nix-config/blob/81275d96047020cd6ed24aecbe5043055c4edf48/modules/nixos/host/persistence.nix#L202-L216), [/persist](https://github.com/DaRacci/nix-config/blob/81275d96047020cd6ed24aecbe5043055c4edf48/modules/nixos/host/persistence.nix#L219-L224) and [/home/racci/Games](https://github.com/DaRacci/nix-config/blob/81275d96047020cd6ed24aecbe5043055c4edf48/modules/home-manager/purpose/gaming/default.nix#L19) (This is proxied to impermanence through [this](https://github.com/DaRacci/nix-config/blob/81275d96047020cd6ed24aecbe5043055c4edf48/modules/home-manager/user/persistence.nix#L102-L117)) 

After creating files with the command `dd if=/dev/zero of=<path>/test.img count=10 bs=1G` on the directly mounted btrfs subvolume, a impermanence mount with the NixOS module and one with the home-manager module, I've run a read speed test and noticed roughly 40-60% performance impact in read and write speeds.

Is this a simple side effect of the mounting method that is used for mounting user directories due to permissions or something, and is there any way to scrape back any of the lost performance via different mounting methods?

Write:
image

Read:
image

Mounts:
image

### LINKED ISSUE #256: [BUG] `fusermount` not found for bindfs -> symlink
**Description:** I recently switched from bindfs to symlinks for some directories. During this transition, I encountered an issue where the `cleanEmptyLinkTargets` activation script is executed before the `unmountPersistentStoragePaths` script, which adds `/run/wrappers/bin` to the PATH. As a result, the `fusermount` command is not found when `cleanEmptyLinkTargets` runs.

**Proposed Solution:** A potential fix could involve adding `/run/wrappers/bin` to the PATH separately, similar to commit cd56321, or restructuring the order of the activation scripts to ensure the PATH is set correctly.

**Workaround:**
Add this to your nix config:
```nix
  home.activation.fixPathForImpermanence = lib.hm.dag.entryBefore ["cleanEmptyLinkTargets"] ''
    PATH=$PATH:/run/wrappers/bin
  '';
```

### LINKED ISSUE #257: [Bisected] "error: expected a set but found a string"
In commit https://github.com/nix-community/impermanence/commit/339078d596713ad32447a99d6e5a59413bda6626 I get the following error when building my configuration

```
> Building NixOS configuration
warning: updating lock file '/home/different/nix-files/flake.lock':
• Updated input 'impermanence':
    'github:nix-community/impermanence/c64bed13b562fc3bb454b48773d4155023ac31b7?narHash=sha256-1Zl9xahw399UiZSJ9Vxs1W4WRFjO1SsNdVZQD4nghz0%3D' (2025-01-12)
  → 'github:nix-community/impermanence/339078d596713ad32447a99d6e5a59413bda6626?narHash=sha256-0q2/VfwClGC88k6V3lIyzbkCeGQi28ASc5puiUDughk%3D' (2025-01-25)
error:
       … while calling the 'head' builtin
         at /nix/store/yw0fhl1cpdi221zhbx2pgi05ykqz636d-source/lib/attrsets.nix:1574:11:
         1573|         || pred here (elemAt values 1) (head values) then
         1574|           head values
             |           ^
         1575|         else

       … while evaluating the attribute 'value'
         at /nix/store/yw0fhl1cpdi221zhbx2pgi05ykqz636d-source/lib/modules.nix:846:9:
          845|     in warnDeprecation opt //
          846|       { value = addErrorContext "while evaluating the option `${showOption loc}':" value;
             |         ^
          847|         inherit (res.defsFinal') highestPrio;

       … while evaluating the option `system.build.toplevel':

       … while evaluating definitions from `/nix/store/yw0fhl1cpdi221zhbx2pgi05ykqz636d-source/nixos/modules/system/activation/top-level.nix':

       … while evaluating the option `home-manager.users.different.home.file':

       … while evaluating definitions from `/nix/store/2fiq1phsggb6llcf2n5hjwscmfzmcwlb-source/home/common/persistence.nix':

       (stack trace truncated; use '--show-trace' to show the full, detailed trace)

       error: expected a set but found a string: ".local/share/nix/repl-history"
```

On the latest commit https://github.com/nix-community/impermanence/commit/4b3e914cdf97a5b536a889e939fb2fd2b043a170 the error is a little different:

```
> Building NixOS configuration
warning: updating lock file '/home/different/nix-files/flake.lock':
• Updated input 'impermanence':
    'github:nix-community/impermanence/339078d596713ad32447a99d6e5a59413bda6626?narHash=sha256-0q2/VfwClGC88k6V3lIyzbkCeGQi28ASc5puiUDughk%3D' (2025-01-25)
  → 'github:nix-community/impermanence/4b3e914cdf97a5b536a889e939fb2fd2b043a170?narHash=sha256-LJggUHbpyeDvNagTUrdhe/pRVp4pnS6wVKALS782gRI%3D' (2025-01-25)
evaluation warning: rustPlatform.fetchCargoTarball is deprecated in favor of rustPlatform.fetchCargoVendor.
                    If you are using buildRustPackage, try setting useFetchCargoVendor = true and regenerating cargoHash.
                    See the 25.05 release notes for more information.
evaluation warning: different profile: `gtk.catppuccin.enable` and `gtk.catppuccin.gnomeShellTheme` are deprecated and will be removed in a future release.

                    The upstream port has been archived and support will no longer be provided.
                    Please see https://github.com/catppuccin/gtk/issues/262
error:
       … while calling the 'head' builtin
         at /nix/store/yw0fhl1cpdi221zhbx2pgi05ykqz636d-source/lib/attrsets.nix:1574:11:
         1573|         || pred here (elemAt values 1) (head values) then
         1574|           head values
             |           ^
         1575|         else

       … while evaluating the attribute 'value'
         at /nix/store/yw0fhl1cpdi221zhbx2pgi05ykqz636d-source/lib/modules.nix:846:9:
          845|     in warnDeprecation opt //
          846|       { value = addErrorContext "while evaluating the option `${showOption loc}':" value;
             |         ^
          847|         inherit (res.defsFinal') highestPrio;

       … while evaluating the option `system.build.toplevel':

       … while evaluating definitions from `/nix/store/yw0fhl1cpdi221zhbx2pgi05ykqz636d-source/nixos/modules/system/activation/top-level.nix':

       … while evaluating the option `system.systemBuilderArgs':

       … while evaluating definitions from `/nix/store/yw0fhl1cpdi221zhbx2pgi05ykqz636d-source/nixos/modules/system/activation/activatable-system.nix':

       … while evaluating the option `system.activationScripts.etc.text':

       … while evaluating definitions from `/nix/store/yw0fhl1cpdi221zhbx2pgi05ykqz636d-source/nixos/modules/system/etc/etc-activation.nix':

       … while evaluating definitions from `/nix/store/yw0fhl1cpdi221zhbx2pgi05ykqz636d-source/nixos/modules/system/etc/etc.nix':

       … while evaluating the option `environment.etc.dbus-1.source':

       (stack trace truncated; use '--show-trace' to show the full, detailed trace)

       error: expected a string but found a set: { directory = ".config/qalculate"; method = "bindfs"; }
```

I can build perfectly on commit https://github.com/nix-community/impermanence/commit/c64bed13b562fc3bb454b48773d4155023ac31b7

talyz:
This PR is a pretty big refactoring and changes a lot of things. When viewing the diffs, I would recommend hiding whitespace changes.

## Replace the current Home Manager module implementation with an interface to the NixOS module

This is the main goal. The other changes, while useful on their own, were mostly done to make this doable.

This should close a whole bunch of issues related to the Home Manager module and massively lessen the maintenance burden of the project, since there's now one unified implementation instead of two disparate ones. This also means that things that were previously not possible, such as using real bind mounts, now are.

The main caveat is that the NixOS module has to be imported into the same configuration as Home Manager for this to work. I can't find any good reason to use the Home Manager module exclusively, but if you have one - do tell.

There are some minor UI and feature differences, mainly on the positive side, but assertions should have been added for any removed or changed functionality.

Fixes #247, fixes #248, fixes #207, fixes #184, fixes #177, fixes #142, fixes #130, fixes #45, fixes #198, fixes #232, fixes #256, fixes #257

## Move all submodule options out to a separate file

This makes it possible to import shared submodule options where needed, reducing duplication and complexity.

## Use `systemd.mounts` instead of `fileSystems` for bind mounts

This removes the need for the `home` option, as the home directory can be deduced automatically. This was previously not possible due to a long-standing issue in nixpkgs: https://github.com/NixOS/nixpkgs/issues/24570.

It also greatly simplifies the `neededForBoot` directory creation, since it's either handled automatically by systemd in the initrd or easily done in the `postMountCommands`.

Fixes #179, fixes #222, fixes #227

## Allow setting file linking method

For some files, it's preferable to use symlinks over bind mounts - for example when a program wants to rename the file temporarily. This adds an option (`method = "symlink"`) to keep a file a symlink even when the file exists, which would otherwise trigger the usage of a bind mount.

Closes #146

## Add basic NixOS tests

Test the basic and critical functionality of the modules. The tests can be run with `nix flake check` and are run automatically in the CI checks. Should have been added a long time ago - I've just been lazy :)

cc @Ma27 @eyJhb @nazarewk @MattSturgeon @colemickens @SrEstegosaurio @DaRacci @mjm @collinarnett @zackattackz

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -16,10 +16,24 @@
       nixosModules.default = self.nixosModules.impermanence;
       nixosModules.impermanence = import ./nixos.nix;
 
-      homeManagerModules.default = self.homeManagerModules.impermanence;
-      homeManagerModules.impermanence = import ./home-manager.nix;
-
       # Deprecated
+      homeManagerModules.default = self.homeManagerModules.impermanence;
+      homeManagerModules.impermanence = {
+        assertions = [
+          {
+            assertion = false;
+            message = ''
+              home.persistence: The Home Manager flake outputs are deprecated!
+
+                The Home Manager module will be automatically imported by the NixOS
+                module. Please remove any manual imports.
+
+                See https://github.com/nix-community/impermanence?tab=readme-ov-file#home-manager
+                for updated usage instructions.
+            '';
+          }
+        ];
+      };
       nixosModule = self.nixosModules.impermanence;
       nixosModules.home-manager.impermanence = self.homeManagerModules.impermanence;
 
@@ -64,6 +78,11 @@
                       users.users.bird = {
                         isNormalUser = true;
                         uid = 1000;
+                      };
+
+                      users.users.fish = {
+                        isNormalUser = true;
+                        uid = 1001;
                       };
 
                       virtualisation.fileSystems = {
@@ -179,28 +198,37 @@
               name = "hm-persistence";
               configuration = { config, ... }:
                 {
-                  home-manager.users.bird.home.stateVersion = config.system.stateVersion;
-
                   imports = [
                     home-manager.nixosModules.home-manager
                   ];
 
-                  home-manager.users.bird = {
-                    imports = [ ./home-manager.nix ];
-                    home.persistence.main = {
-                      persistentStoragePath = "/persistent";
-                      directories = [
-                        "Downloads"
-                        "Music"
-                        "Pictures"
-                        "Documents"
-                        "Videos"
-                      ];
-                      files = [
-                        ".config/persistence_test"
-                      ];
+                  home-manager.sharedModules = [{ home.stateVersion = config.system.stateVersion; }];
+
+                  home-manager.users.bird =
+                    {
+                      home.persistence.main = {
+                        persistentStoragePath = "/persistent";
+                        directories = [
+                          "Downloads"
+                          "Music"
+                          "Pictures"
+                          "Documents"
+                          "Videos"
+                        ];
+                        files = [
+                          ".config/persistence_test"
+                        ];
+                      };
                     };
-                  };
+
+                  home-manager.users.fish =
+                    {
+                      home.file = {
+                        "useless".text = ''
+                          a useless file
+                        '';
+                      };
+                    };
                 };
             };
           }
--- a/home-manager.nix
+++ b/home-manager.nix
@@ -1,7 +1,6 @@
 { pkgs
 , config
 , lib
-, persistenceModuleImported
 , ...
 }:
 
@@ -18,6 +17,7 @@ let
   inherit (types)
     attrsOf
     submodule
+    bool
     ;
 
   inherit (config) home;
@@ -27,48 +27,48 @@ let
   persistentStoragePaths = catAttrs "persistentStoragePath" (attrValues cfg);
 in
 {
-  options.home.persistence = mkOption {
-    default = { };
-    type = attrsOf (
-      submodule (
-        { name, config, ... }:
-        import ./submodule-options.nix {
-          inherit pkgs lib name config;
+  options =
+    {
+      home.persistence = mkOption {
+        default = { };
+        type = attrsOf (
+          submodule (
+            { name, config, ... }:
+            import ./submodule-options.nix {
+              inherit pkgs lib name config;
 
-          user = home.username;
-          homeDir = home.homeDirectory;
+              user = home.username;
+              homeDir = home.homeDirectory;
 
-          # Home Manager doesn't seem to know about the user's group,
-          # so we default it to null here and fill it in in the NixOS
-          # module instead
-          group = null;
-        }
-      ));
-  };
-  config = {
-    _module.args = {
-      persistenceModuleImported = false;
+              # Home Manager doesn't seem to know about the user's group,
+              # so we default it to null here and fill it in in the NixOS
+              # module instead
+              group = null;
+            }
+          ));
+      };
+      home._nixosModuleImported = mkOption {
+        default = false;
+        type = bool;
+        internal = true;
+        description = ''
+          Internal option to signal whether the NixOS persistence
+          module was properly imported. Do not set this!
+        '';
+      };
     };
+  config = {
     assertions = [
       {
-        assertion = config.submoduleSupport.enable;
+        assertion = config.home._nixosModuleImported;
         message = ''
-          home.persistence: Home Manager used standalone!
+          home.persistence: Module was imported manually!
 
-            Home Manager has to be imported as a module in your NixOS
-            configuration for the persistence module to work properly. See
-            https://nix-community.github.io/home-manager/#sec-install-nixos-module
-            for instructions.
-        '';
-      }
-      {
-        assertion = persistenceModuleImported;
-        message = ''
-          home.persistence: NixOS persistence module missing!
-
-            The Home Manager module requires the NixOS module to work properly. See
-            https://github.com/nix-community/impermanence?tab=readme-ov-file#nixos
-            for instructions.
+            The Home Manager persistence module should not be imported
+            manually. It will be imported by the NixOS module
+            automatically. See
+            https://github.com/nix-community/impermanence?tab=readme-ov-file#home-manager
+            for instructions and examples.
         '';
       }
       {
--- a/nixos.nix
+++ b/nixos.nix
@@ -221,8 +221,13 @@ in
 
   config =
     mkMerge [
-      (lib.optionalAttrs (options ? home-manager.extraSpecialArgs) {
-        home-manager.extraSpecialArgs.persistenceModuleImported = true;
+      (lib.optionalAttrs (options ? home-manager.sharedModules) {
+        home-manager.sharedModules = [
+          ./home-manager.nix
+          {
+            home._nixosModuleImported = true;
+          }
+        ];
       })
       (mkIf (allPersistentStoragePaths != { })
         (mkMerge [
--- a/README.org
+++ b/README.org
@@ -340,28 +340,14 @@
     the ~nixos~ directory of Home Manager's repo) *and* the NixOS
     ~persistence~ module in order for this module to work as intended.
 
-    To use the module, import it into your configuration with
+    The module will be loaded automatically when you [[*NixOS][import the NixOS
+    persistence module]] and the Home Manager NixOS module.
 
-    #+begin_src nix
-      {
-        imports = [
-          /path/to/home-manager/nixos
-          /path/to/impermanence/nixos.nix
-        ];
+    It adds the ~home.persistence~ option, which is an attribute set
+    of submodules, where the attribute name is the path to persistent
+    storage.
 
-        home-manager.users.bird =
-          { ... }:
-          {
-            imports = [
-              /path/to/impermanence/home-manager.nix
-              ./home/config.nix # The rest of your Home Manager
-                                # configuration
-            ];
-          };
-      }
-    #+end_src
-
-    or use the provided ~homeManagerModules.impermanence~ flake output:
+    Usage is shown best with an example:
 
     #+begin_src nix
       {
@@ -388,47 +374,30 @@
                   ];
 
                   home-manager.users.bird =
-                    { ... }:
                     {
-                      imports = [
-                        impermanence.homeManagerModules.impermanence
-                        ./home/config.nix # The rest of your Home Manager
-                                          # configuration
-                      ];
-                    };
+                      home.persistence."/persistent" = {
+                        directories = [
+                          "Downloads"
+                          "Music"
+                          "Pictures"
+                          "Documents"
+                          "Videos"
+                          "VirtualBox VMs"
+                          { directory = ".gnupg"; mode = "0700"; }
+                          { directory = ".ssh"; mode = "0700"; }
+                          { directory = ".nixops"; mode = "0700"; }
+                          { directory = ".local/share/keyrings"; mode = "0700"; }
+                          ".local/share/direnv"
+                        ];
+                        files = [
+                          ".screenrc"
+                        ];
+                      };
+                  };
                 }
               ];
             };
           };
-      }
-    #+end_src
-
-    This adds the ~home.persistence~ option, which is an attribute set
-    of submodules, where the attribute name is the path to persistent
-    storage.
-
-    Usage is shown best with an example:
-
-    #+begin_src nix
-      {
-        home.persistence."/persistent" = {
-          directories = [
-            "Downloads"
-            "Music"
-            "Pictures"
-            "Documents"
-            "Videos"
-            "VirtualBox VMs"
-            { directory = ".gnupg"; mode = "0700"; }
-            { directory = ".ssh"; mode = "0700"; }
-            { directory = ".nixops"; mode = "0700"; }
-            { directory = ".local/share/keyrings"; mode = "0700"; }
-            ".local/share/direnv"
-          ];
-          files = [
-            ".screenrc"
-          ];
-        };
       }
     #+end_src
 


mount-file.bash
```diff
@@ -10,14 +10,15 @@ shopt -s inherit_errexit  # Inherit the errexit option status in subshells.
 trap 'echo Error when executing ${BASH_COMMAND} at line ${LINENO}! >&2' ERR
 
 # Get inputs from command line arguments
-if [[ $# != 3 ]]; then
+if [[ $# != 4 ]]; then
     echo "Error: 'mount-file.bash' requires *three* args." >&2
```
DaRacci, line 14: Should read `*four* args`, however may be worth documenting the argument names/description in the error.

DaRacci:
From what you've described it sounds a lot like what i transformed my impermanence setup to be like a while back (but much more fleshed out); I made [change](https://github.com/DaRacci/nix-config/commit/cbf785ba5eba4e3098f6e62804ae13362ee72574#diff-1311bb772737b843e5c4070455b6371c977a5a553d048dae2b19afa1ed75269f) that just redirected all my home declarations for it to the NixOS module to avoid bindfs or symlinks.

Definitely a major improvement with regards to performance.

It may be worth trying to include a notice, or a cleanup for BindFS mounts to prevent any issues during the switch over as I imagine #249 may rear its head out; This change also cause an issue when attempting to live switch a configuration due to in use mounts but that's likely unsolvable.

mount-file.bash
talyz: Ah, yes, of course! Well, it's just a sanity check. The script is not really supposed to be used interactively, so this is just to catch a situation where we're not feeding it all the required arguments programmatically.

Ma27:
I think it's worth mentioning that the unit-based approach was pioneered by @WilliButz already with https://github.com/nix-community/preservation.

Perhaps an exchange of experiences could help?

different-name:
Just got this set up on my machine

> Replace the current Home Manager module implementation with an interface to the NixOS module

I made a small module to track stray files in persisted directories and files that haven't yet been persisted, and having everything now be in the same format and structure now is very helpful!

> assertions should have been added for any removed or changed functionality.

The migration to v2 was very smooth thanks to these assertions

I haven't encountered any problems so far, I use the impermanence module across my entire system

Thank you for all the hard work @talyz :heart: I have been using impermanence since I first started using NixOS a year ago

flake.nix
```diff
@@ -1,13 +1,209 @@
 {
-  outputs = { self }: {
-    nixosModules.default = self.nixosModules.impermanence;
-    nixosModules.impermanence = import ./nixos.nix;
+  inputs = {
+    nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable";
+    home-manager.url = "github:nix-community/home-manager";
+    home-manager.inputs.nixpkgs.follows = "nixpkgs";
+  };
```
MattSturgeon, line 6: These inputs will be propagated to consuming flakes, unless users take steps to have these inputs "follow" their own (or follow `""`).

Currently, all inputs that end up in the user's flake.lock will get downloaded.

Since these inputs are only used by dev outputs (`devShells` and `checks`) I wonder if it'd be better to have either a separate "dev" flake or use non-flake nix files for the devshell and checks?

This would mean dev stuff continues to work without the "public" flake having additional dependencies.

nixos.nix
```diff
@@ -4,454 +4,184 @@ let
...
+        let
+          paths = flatten
+            (mapAttrsToList
+              (_name: value:
+                attrValues (value.home.persistence or { }))
+              config.home-manager.users or { });
+        in
+        filter (v: v.enable) paths;
+    in
+    zipAttrsWith (_: v: flatten v) (nixos ++ nixosUsers ++ homeManager);
```
MattSturgeon, line 79:
```suggestion
    zipAttrsWith (_: flatten) (nixos ++ nixosUsers ++ homeManager);
```

Could this use [`concatLists`](https://noogle.dev/f/builtins/concatLists) instead of `flatten`?:

```suggestion
    zipAttrsWith (_: concatListsn) (nixos ++ nixosUsers ++ homeManager);
```

nixos.nix
```diff
@@ -4,454 +4,184 @@ let
...
-    };
-  };
-
-  # Create all fileSystems bind mount entries for a specific
-  # persistent storage path.
-  bindMounts = listToAttrs (map mkBindMountNameValuePair directories);
 in
 {
   options = {
+    home-manager.extraSpecialArgs = mkOption { };
```
MattSturgeon, line 111: I assume you're declaring this option because you depend on it existing?

Would it be better to check `options ? home-manager.extraSpecialArgs` instead, wherever you use/read/set this option?

E.g.

```nix
mkMerge [
  (optionalAttrs (options ? home-manager.extraSpecialArgs) {
    home-manager.extraSpecialArgs.persistenceModuleImported = true;
  })
```

Alternatively, if it doesn't need to be a "special" arg, you could instead define `_module.args` in the home-manager submodule; this is usually preferred over using `specialArgs` anyway.

```nix
someHomeManagerModule = {
  _module.args.persistenceModuleImported = true;
}
```

Of course, this won't work if you're using `persistenceModuleImported` to decide which modules to import.

nixos.nix
```diff
@@ -4,454 +4,184 @@ let
...
-                          x)
-                      dirOpts.options;
-                  filePath = mkOption {
-                    type = path;
-                    internal = true;
-                  };
-                };
-              };
-              dirOpts = {
+            recursiveUpdate
```
MattSturgeon, line 119: Why use `recursiveUpdate` when you can just have multiple modules in this submodule?

Remember, a submodule is its own fully-blown module eval, so it can have as many modules as you like. Just tweak `submodule-options.nix` so that it returns a full module instead of a set of options.

If you need to pass arbitrary stuff into a module, use [`importApply`](https://noogle.dev/f/lib/modules/importApply):

```nix
imports = [
  (lib.modules.importApply ./submodule-options.nix {
    inherit pkgs;
    user = "root";
    group = "root";
    homeDir = null;
  })
```

NOTE: `lib` `name` & `config`\* can be accessed as module args within the submodule eval, however `pkgs` comes from the outer module eval so needs to be passed in.  
\*Of course this is the submodule's `config`, not the outer `config`.

nixos.nix
```diff
@@ -4,454 +4,184 @@ let
...
+                    in
+                    mkOption {
+                      type = attrsOf (
+                        submodule (
+                          { name, config, ... }:
+                          import ./submodule-options.nix {
+                            inherit pkgs lib;
+                            config = outerConfig // config;
+                            name = outerName;
+                            usersOpts = true;
```
MattSturgeon, line 141: Again, this feels like it's fighting the module system.

I'm not too keen on `config = outerConfig // config` either, surely it's better to be explicit about those being separate?

submodule-options.nix
```diff
@@ -0,0 +1,323 @@
+{ pkgs
+, lib
+, name
+, config
+, homeDir
+, usersOpts ? false  # Are the options used as users.<username> submodule options?
+, user               # Default user name
+, group              # Default user group
+}:
```
MattSturgeon, line 9: If going with my suggestion to promote this to a "real" option declaring module, you'd split these args something like:

```suggestion
# Usage: lib.modules.importApply ./submodule-options.nix { /* args */ }

# Applied args
{ pkgs
, outerConfig ? null
, homeDir
, usersOpts ? false  # Are the options used as users.<username> submodule options?
, user               # Default user name
, group              # Default user group
}:

# (Sub)module args
{ lib
, name
, config
}:
```

The rest of this file is already a valid module and shouldn't need any changes, other than to handle `outerConfig` as needed.

MattSturgeon (commented):
I'm not very familiar with the inner workings of impermanence, but I thought I'd give the diff a quick once-over to see if I could help with any small suggestions :slightly_smiling_face:

flake.nix
cardboardcpu: Or maybe it could be documented in the setup page (or wherever else) that a user might `.follows = ""` since they are dev dependencies.

I have seen a few projects do that and it does not requiere any changes in the flake or workflow.

MattSturgeon: Yes, this is a valid approach.

But it'd be a _nicer_ user-experience if it was solved here instead of users needing to solve it themselves (even with documented instructions).

Not a major issue, of course.

home-manager.nix
```diff
@@ -1,486 +1,86 @@
...
+          # module instead
+          group = null;
         }
-      '';
-    };
-
+      ));
   };
-
   config = {
```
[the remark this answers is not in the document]
gigamonster256: This is exactly my use case mentioned [above](https://github.com/nix-community/impermanence/pull/272#issuecomment-2937716517) and is the same as the fix I applied and have been using [in my own branch](https://github.com/gigamonster256/impermanence/commit/8f17f576cfc1c4b0d2663972389f26759a4b127c)

different-name:
I've been using this daily for a bit over a month now, I only have two issues with the new method

- Applications like Thunar and VSC don't seem to support trashing on system internal mounts, though tools like trashy and trash-cli seem to handle it fine
- `lsblk` is quite cluttered with mount points

Other than that, I have no issues!

mtroberts:
I'm trying to test this out, working through the assertions. I'm getting the error:

```
 Failed assertions:
       - The option definition `removePrefixDirectory' in `/nix/store/c0sdj9ibra25kd732rrr5wpffwy6y2pb-source/modules/nixosModules/system/impermanence.nix' no longer has any effect; please remove it.
       ▹ persistence."/cache/hm":
           The use of prefix directories is deprecated and the functionality has been removed.
           If you depend on this functionality, use the `home-manager-v1' branch.
```

I don't have that option set anywhere in my config. I've never used it.

mtroberts:
> I'm trying to test this out, working through the assertions. I'm getting the error:
> 
> ```
>  Failed assertions:
>        - The option definition `removePrefixDirectory' in `/nix/store/c0sdj9ibra25kd732rrr5wpffwy6y2pb-source/modules/nixosModules/system/impermanence.nix' no longer has any effect; please remove it.
>        ▹ persistence."/cache/hm":
>            The use of prefix directories is deprecated and the functionality has been removed.
>            If you depend on this functionality, use the `home-manager-v1' branch.
> ```
> 
> I don't have that option set anywhere in my config. I've never used it.

Never mind, I forgot I was using Geometer1729/persist-retro.

mtroberts:
> ```
> Now I'm getting this error relating to Plasma-Manager: 
> ```

It looks like my intermediate files (for example .config) are getting created with root permissions. This means anything that tries to create a folder in say, .config or .local don't have permission. Any ideas?

cardboardcpu:
I cannot seam to even eval the PR, it fails giving me this error:
```
❯ : nix flake check github:nix-community/impermanence/home-manager-v2
error: attribute 'home' already defined at /nix/store/pr2nf9bp6fz91vdxii9pv8c4larp1drb-source/flake.nix:190:21
       at /nix/store/pr2nf9bp6fz91vdxii9pv8c4larp1drb-source/flake.nix:182:19:
          181|                 {
          182|                   home-manager.users.bird.home.stateVersion = config.system.stateVersion;
             |                   ^
          183|
```

What I am doing wrong?

marrobHD:
I switched the repo in the flake.nix
Now it fails with assertions. I don't get why it fails. Home Manager is imported in the flake.
```
       error:
       Failed assertions:
       - home.persistence: Home Manager used standalone!

         Home Manager has to be imported as a module in your NixOS
         configuration for the persistence module to work properly. See
         https://nix-community.github.io/home-manager/#sec-install-nixos-module
         for instructions.

       - home.persistence: NixOS persistence module missing!

         The Home Manager module requires the NixOS module to work properly. See
         https://github.com/nix-community/impermanence?tab=readme-ov-file#nixos
         for instructions.
```

flake.nix
talyz: This feels like a flake issue (and there is at least [one open for it](https://github.com/NixOS/nix/issues/6124). I think it's fine to have the consumer set `.follows = ""` if they care about this - at least until there's a better solution.

nixos.nix
MattSturgeon: > The argument has to be fed to the home manager module from NixOS, not the other way around.

Sure, but module args (`specialArgs` and `_module.args`) are not the only ways to get a value from an outer module configuration into a sub-module configuration.

You can also simply define something in [`home-manager.sharedModules`](https://nix-community.github.io/home-manager/nixos-options.xhtml#nixos-opt-home-manager.sharedModules), e.g. defining some internal option with the value.

From my perspective, using args generally, and specialArgs in particular, should be a last resort.

flake.nix
MattSturgeon: It is definitely a flakes issue, but one that we're stuck with for the foreseeable future.

When using flake-parts, one can use ["flake partitions"](https://flake.parts/options/flake-parts-partitions.html) to have some inputs only fetched when using the parts of the flake that need them.

When not using flake-parts, that's still possible with some manual effort.

As it's not a major issue, it can be handled separately if anyone is motivated enough. I only brought it up now as this is the first case of dev-inputs being added to the public flake's inputs.

flake.nix
```diff
@@ -1,13 +1,209 @@
...
+              name = "hm-persistence";
+              configuration = { config, ... }:
+                {
+                  home-manager.users.bird.home.stateVersion = config.system.stateVersion;
+
+                  imports = [
+                    home-manager.nixosModules.home-manager
+                  ];
+
+                  home-manager.users.bird = {
```
bitbloxhub, line 188: I don't see why we need `home-manager.users.bird.home.stateVersion = config.system.stateVersion;` instead of just putting `home.stateVersion = config.system.stateVersion;` under `home-manager.users.bird`. I think this is cleaner and works with Lix.

flake.nix
```diff
@@ -1,13 +1,209 @@
...
+              };
+          }
+        );
+
+      checks = forAllSystems
+        (system:
+          let
+            pkgs = nixpkgs.legacyPackages.${system};
+            mkTest = { name, configuration }:
+              pkgs.nixosTest {
```
bitbloxhub, line 46: `pkgs.nixosTest` has been mostly replaced with `pkgs.testers.runNixOSTest`, see https://github.com/NixOS/nixpkgs/pull/293891, https://github.com/NixOS/nixpkgs/pull/292759, and https://discourse.nixos.org/t/nixostest-vs-runtest/26419/5.
```suggestion
              pkgs.testers.runNixOSTest {
```

bitbloxhub (commented):
Didn't go through the whole thing, but i got it to eval with Lix and found another small nitpick.

flake.nix
unhitched9271: is this change likely to be made? I'd start test driving this with bcachefs (which I assume would be useful), but unfortunately this is blocking me doing that (as I also use Lix)

clhodapp:
> The main caveat is that the NixOS module has to be imported into the same configuration as Home Manager for this to work. I can't find any good reason to use the Home Manager module exclusively, but if you have one - do tell.

Just in general, it's icky to have to become root to update what is ostensibly your user-level configuration. This has always been my issue with the idea of the home-manager / NixOS integration. I'm perfectly happy to make my current NixOS config available to home-manager, but I don't like NixOS system rebuilds being the rollout mechanism for home-manager configurations.

nixos.nix
talyz: Basically because I forgot that this was an option when restructuring :) Since `submodule-options.nix` already was a full-blown module, this was easy to address.

flake.nix
talyz: Thanks!

talyz:
> Just in general, it's icky to have to become root to update what is ostensibly your user-level configuration. This has always been my issue with the idea of the home-manager / NixOS integration.

Yeah, if it was possible to do it without requiring it to be loaded as a NixOS module, that would of course be preferable. However, this was already a requirement in v1 due to activation timing issues, where the users' activation scripts have to run before the users log in.

talyz:
> I switched the repo in the flake.nix Now it fails with assertions. I don't get why it fails. Home Manager is imported in the flake.

More context would be necessary to figure this out. Did you use `home-assistant switch` or does this happen when you rebuild your NixOS configuration? Could you try the latest version?

flake.nix
CathalMullan: Can confirm this works fine with Lix now.

talyz:
> The old user unit approach was nice because the mounts wouldn't start until the user logged in.

If used as intended, it would. The system `home-manager-<user>.service` unit should start the bind mounts before the user logs in and this has been the only officially supported scenario.

Feel free to work on a way to handle this, though! It definitely sounds like a nice feature to have.

philipkbh:
Weird because under v1 I don't run into this issue - there all parent folders belong to my user and not to root.

somasis:
> Weird because under v1 I don't run into this issue - there all parent folders belong to my user and not to root.

Also having this issue, unfortunately. Having to use tmpfiles to fix it kinda feels like a regression in my opinion...

raj-magesh:
@philipkbh @somasis @upidapi I've created a new issue highlighting this breakage.

IMO creating parent directories with the right ownership (or copying over the ownership from the existing directories in persistent storage) should be automated by impermanence by default and not left to the user to do manually.

This issue didn't happen with V1.

Issue #298, opened 2026-01-15
Parent directories are owned by `root/root` instead of `<user>/users` when using `home.persistence`


raj-magesh:
Hi there!

Thanks for the work on V2: I'm sure it's cleaned up a lot of the internals and made maintenance easier!

However, there's a pretty important regression when using `home.persistence` options: parent directories are created with owner and group as root when they shouldn't be, and this messes up permissions and prevents the files from being bind-mounted.

As an example, when using the following configuration, `.local` and `.local/share` are both created with owner/group `root/root` instead of `my_user/users`, even when the `user` and `group` options are set explicitly.

```
  home.persistence."/persist" = {
    directories = [ 
      {
        directory = ".local/share/nvim";
        user = "my_user";
        group = "users";
      }
    ];
  };
```

Note that `/persist/home/my_user/.local` and `/persist/home/my_user/.local/share` exist already and have the correct ownership (`my_user/user`).

This also happens when persisting files, and even when the `parentDirectory.{user, group}` options are set explicitly.

This issue was reported by these users in the V2 PR (https://github.com/nix-community/impermanence/pull/272#issuecomment-3154492310, https://github.com/nix-community/impermanence/pull/272#issuecomment-3733574385, https://github.com/nix-community/impermanence/pull/272#issuecomment-3736578986). A potential reason for this was identified by this commenter (https://github.com/nix-community/impermanence/pull/272#issuecomment-3733635431) along with a possible solution using `systemd.tmpfiles` but that seems very inelegant for the user to have to do that manually for the entire parent directory tree, especially when it can get pretty deep. If necessary, I think it should be abstracted over by impermanence automatically.

Vanadium5000:
I have the same issue

talyz:
I have not been able to reproduce this. I added a parent directory ownership check to the tests and it passes.

The suggested issues or solutions are invalid. `mkdir` is never called with `-p` and #302 breaks the default permission assignment.

My guess at this point would be that either `home.username` is somehow incorrectly set or there's a race condition where the directory creation script runs after the mount units start, but I can only speculate.

A complete journald log from a boot where this goes wrong and with `environment.persistence.<name>.enableDebugging = true;` would be very helpful. Also, please make sure `home.username` gets set to the correct user name.

AWildLeon:
I could reproduce this.
it seems to be a race condition.
its fixed if i set ``fileSystems."/home".neededForBoot = true;``

talyz:
[...]

PR #302, opened 2026-01-16
Fix: Parent directories are owned by root/root instead of <user>/users when using home.persistence

Vanadium5000:
Basically seems to fix issue: https://github.com/nix-community/impermanence/issues/298, which makes a load of apps, etc break because folders like .cache, .local, and whatever contains persisted files/folders in home becomes owned by root. This is a very annoying issue as every time I rebuild, it makes .local, etc owned by root.


Vanadium5000:
@talyz Tested it, it works on reboot

PR #304, opened 2026-01-18
regression: Do not treat `/` as a "home"

glaeqen:
`parentDirectory.home` was set to `/` for non-HM managed files which caused `/` to be given 0700 permissions. This made certain systemd units running as non-root uncapable of accessing the filesystem and at worst rendering the machine unbootable.

### DIFF
--- a/submodule-options.nix
+++ b/submodule-options.nix
@@ -187,16 +187,16 @@ let
     }
     ({ config, ... }:
       let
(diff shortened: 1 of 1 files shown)


glaeqen:
I'm not sure how does the test suite didn't catch this or that no one caught this in normal usage because even the `/etc/machine-id` from README is causing the `0700 on /` issue. Am I missing something?

PR #309, opened 2026-02-14
Feature: home-relative persistent storage roots

### LINKED ISSUE #296: Per directory disabling automatic /home/username suffix addition in home.persistence
I had separate drive mounted to /persist/data and i wanted it to be available to every user via ~/data so i written:
```nix
# before update
  home.persistence."/persist" = {
    directories = ["data"];
  };
```
but after this update i needed to convert all my other persists `home.persistence."/persist/home/..."`  to  `home.persistence."/persist"` so now they are in the same list and:
```nix
  home.persistence."/persist" = {
    directories = [
      {
        directory = "data";
        persistentStoragePath = "/persist";
      }
      "Desktop"
      "Documents"
      "Downloads"
      ...
```
looks like a solution but it's not working? 

question: how to disable automatic addition of `/home/username` suffix to persistent storage paths on single directories entry when using home.persistence?

ctem:
Fixes #296

## Background

Prior to the Home Manager v2 update (which is a work of art btw), persistent storage path (PSP) roots were relative to their modules' respective domains. That is:

- NixOS module PSP roots were system root-relative.
- HM module PSP roots were user home-relative.

## Issue

Now that the HM module is an interface to the NixOS module, all PSP roots are system root-relative. This requires home-relative PSPs (e.g., typical user home "dotfile repos") to be restructured such that all content is moved into a host-specific user home (e.g., from `dotfiles_repo/` into `dotfiles_repo/home/bird/`), which is a dealbreaker for repos that store "home"-scoped data that cannot be tied to a single home directory parent or user name.

## Proposed Solution

This PR introduces an option, `homeRelative`, to toggle whether the persistent storage path represents the system root (the default) or the user home.


nixos.nix
Musholic, line 264: I think this should be changed to:
```
                    "persist-${escapeSystemdPath mountPoint}" = {
```
Since the targetFile can now be the same value for different users, the mountpoint should be better at uniquely distinguishing the services.

Musholic, line 274: Same here, this should probably be changed to:

```
                        ExecStop = pkgs.writeShellScript "unbindOrUnlink-${escapeSystemdPath mountPoint}" ''
```

Musholic:
I had a good experience with this PR. It would be great if it could be merged :slightly_smiling_face:. The only other issue (apart from my two review comments) I noticed is that it still creates the ./home/USER directories (owned by root) below the  given persistent storage path

PR #312, opened 2026-02-26
feat: add stub home-manager module for systems without impermanence

### LINKED ISSUE #292: Add shim for homeManager config reuse on systems without impermanence.
I missed the fact that the independent homeManager config was always considered a bug, not a feature.   Still this exposes something that does more broadly affect people.  Keeping the original message below.

~~I've been quite a bit out of the loop on the V2 project, but just ran into an issue today when doing an update:~~
```
error:
       … while evaluating the attribute '"nina@sarah".config'
         at /nix/store/qggvk0pyawzv6mb3pxyi466520pjdd5a-source/modules/default.nix:61:32:
           60|     {
           61|       inherit (module) options config;
             |                                ^
           62|

       … while evaluating a branch condition
         at /nix/store/l04691r5xf6gj386nd8qqigdk29is7yd-source/lib/lists.nix:142:18:
          141|       len = length list;
          142|       fold' = n: if n == len then nul else op (elemAt list n) (fold' (n + 1));
             |                  ^
          143|     in

       (stack trace truncated; use '--show-trace' to show the full, detailed trace)

       error: The option `home.persistence' does not exist. Definition values:
       - In `/nix/store/zswnfdl2iczs61k8dbdxh8rhlwdfz2sp-source/homeManager-modules/desktop/persist.nix':
           {
             _type = "if";
             condition = true;
             content = {
               "/persist/home/nina" = {
           ...
```

~~I'm guessing based on a quick look through that using impermanence on a separated home-manager is not an option anymore? My current config uses a folder inside the persistence directory to store the home-manager config for a user, and activates that during system activation, allowing me to swap out the activation after a system build.  I ended up moving to this as a full system rebuild + home-manager rebuild was beginning to get overly time-consuming (I sadly have resource constrained systems rn, so it's inevitable), and broke it in two to allow the home-manager config to happen a bit quicker, since that was where I was doing most of my regular changes.~~

~~of note, cause of how I handle my configs (it's complicated, but I effectively pushed myself to do a whole stack rebuild instead of individual systems for most stuff so I don't end up with inconsistancies between systems), doing nixos+home-manager every time results in a jump from an average of 30s (after store warmup) to nearly an hour, as each home-manager build must be done for each system, and building 7 systems with between 1 and 4 users each adds way too much time.~~

~~I know I had to add an escape hatch for v1 to allow me to just pack activationPackage myself, but is there any similar escape hatch on v2 to allow me to do the same/similar thing? or is the nixosModule now heavily ingrained into the homeModule?~~

### LINKED ISSUE #310: Share config between NixOS and home-manager?
Ever since https://github.com/nix-community/impermanence/pull/272, the home-manager module can no longer be imported separately.

But if you've previously structured your code like this:

`nix-config/home/vesktop.nix`

```nix
{ pkgs, ... }:
{
  home = {
    packages = with pkgs; [ vesktop ];

    persistence."/persist".directories = [
      ".config/vesktop/sessionData"
      ".config/vesktop/settings"
    ];
  };
}
```

And you share this code between your NixOS+home-manager and (standalone) home-manager configuration, how do you handle the non-existent `home.persistence` in your home-manager config?

Are you expected to have a strict separation between impermanence and the rest of your config?

I think a better approach here is to re-introduce the standalone home-manager module, but make it explicitly no-op.

bitbloxhub:
Fixes https://github.com/nix-community/impermanence/issues/310
Fixes https://github.com/nix-community/impermanence/issues/292

Still need to test this out.


bitbloxhub:
Actually, thinking a bit more about this, could you just set `home._nixosModuleImported` and do `import "${inputs.impermanence}/home-manager.nix"`? I suppose I should try that.

bitbloxhub:
Turns out that solution works fine. Closing.

PR #321, opened 2026-04-20
README: Update example for systemd stage 1

ElvishJerricco:
https://github.com/NixOS/nixpkgs/pull/435781


README.org
```diff
@@ -101,37 +101,60 @@
...
-          }
+        boot.initrd.systemd = {
+          enable = true; # Default in 26.05
+          services.wipe-file-systems = {
+            # Specify dependencies explicitly
+            unitConfig.DefaultDependencies = false;
+            # The script needs to run to completion before this service is done
+            serviceConfig.Type = "oneshot";
+            # This service is required for boot to succeed
+            requiredBy = [ "initrd.target" ];
```
fabiancholewinski1234, line 120: Thinking about `wantedBy` vs `requiredBy`.  The former is a weak dependency, i.e. if the rollback fails, the system still continues the boot process.
The latter is a hard dependency, i.e. the boot will fail.

In the latter case (`requiredBy`), this means a live system or similar is needed to fix the issue.

In the former case (`wantedBy`), the system starts up, and we end up with a possibly non-pristine  filesystem (like "normal setups"). This allows fixing the issue in our "usual" environment (e.g. `nixos-rebuild`) and is thus a smoother experience for the user.

So maybe `wantedBy` is the better dependency type?  (the device dependency should still be `requires` as the rollback will fail if it's not available.) "Non serious" failure conditions like syntax errors in the script (e.g. incorporating some suggestion from issues here), changed partition "name"/"location" (sda1 -> sda2, ...) could be better fixed. While "serious"failure conditions like filesystem corruption or hardware failure still fail at a later stage.

@ElvishJerricco what are your thoughts?

README.org
```diff
@@ -101,37 +101,60 @@
...
+              delete_subvolume_recursively() {
+                  IFS=$'\n'
+                  for i in $(btrfs subvolume list -o "$1" | cut -f 9- -d ' '); do
+                      delete_subvolume_recursively "/btrfs_tmp/$i"
+                  done
+                  btrfs subvolume delete "$1"
+              }
+
+              for i in $(find /btrfs_tmp/old_roots/ -maxdepth 1 -mtime +30); do
+                  delete_subvolume_recursively "$i"
```
fabiancholewinski1234, line 150: I suggest using `btrfs subvolume delete --recursive "$i"` instead of `delete_subvolume_recursively "$i"`.

The `--recursive`/`-R` argument was introduced in *btrfs-progs 6.12*, which hit stable NixOS a long time ago; current stable (nixos-25.11 as of this writing) has *btrfs-progs 6.17.1*.

See #271, which adds this as a comment suggestion. That PR is a year old, where *btrfs-progs 6.12* might not have been available everywhere, but today I would say "use it instead of rolling your own".

ElvishJerricco: Purely for the sake of moving this PR forward as quickly as possible, I think we should leave improvements to the script itself for other PRs. As is, the script is the same as it was in the example before.

README.org
ElvishJerricco: I don't have a strong opinion on which one should be in the example, but it would probably be a good idea to call out the distinction in the comment so the reader can decide for themselves.

I'm realizing the old `postResumeCommands` did not fail the boot when a command would fail. So `wantedBy` matches the old behavior better. Though, it also comes with the caveat that the `script` here in systemd stage 1 will stop at the first failing command, vs `postResumeCommands` proceeding through a failed command onto the following ones. So maybe I should also make the script a little better at cleaning up after itself (specifically, unmounting the mounted FS) in the failure case with an `EXIT` handler.

fabiancholewinski1234: > I'm realizing the old `postResumeCommands` did not fail the boot when a command would fail. So `wantedBy` matches the old behavior better. Though, it also comes with the caveat that the script here in systemd stage 1 will stop at the first failing command, vs `postResumeCommands` proceeding through a failed command onto the following ones.

Didn't know that.  Just looked it up and it seems that the caveat comes with a caveat itself:
It only aborts if `enableStrictShellChecks` is false.

https://github.com/NixOS/nixpkgs/blob/a4bf06618f0b5ee50f14ed8f0da77d34ecc19160/nixos/lib/systemd-lib.nix#L552-L562

https://github.com/NixOS/nixpkgs/blob/3b0bc3fc4c7d3be9b9c2c00bc2598b8b91371ce1/pkgs/build-support/trivial-builders/default.nix#L227-L236

---

> So maybe I should also make the script a little better at cleaning up after itself (specifically, unmounting the mounted FS) in the failure case with an EXIT handler.

FWIW I have this:
```bash
    # ...
    (
      MNTPOINT=$(mktemp -d)
      mount -t btrfs -o subvol=/ ${device} "$MNTPOINT"
      trap 'umount "$MNTPOINT"' EXIT
      # mv current subvolume to old
      # and delete old subvolumes
    )
```
The subshell might be unnecessary but I like to keep my traps as local as possible.

PR #322, opened 2026-04-22
README: add BTRFS for unstable example

### LINKED ISSUE #320: Failed assertion from 'boot.initrd.postResumeCommands'
Hello, Ive been using a btfs impermanence setup like whats shown in the readme for a few years now, but recently I tried to update my flake and got the follow error 
```

error:
       Failed assertions:
       - systemd stage 1 does not support `boot.initrd.postResumeCommands`. Instead, create systemd services using the `boot.initrd.systemd.services` options, which has an API matching the stage 2 `systemd.services` options. Refer to `bootup(7)`, specifically the sections on "Bootup in the Initrd" and "System Manager Bootup", for information about when various units happen, and order services accordingly.

           Definitions:
           - /nix/store/94mak8iycx0l1clnnn1f0dxfg5l8gplq-source/hosts/Titanic/configuration.nix
Command 'nix --extra-experimental-features 'nix-command flakes' build --print-out-paths '/persist/NixOSDots#nixosConfigurations."Titanic".config.system.build.toplevel' --no-link' returned non-zero exit status 1.
[ble: exit 1]
```

Im afraid im not knowledgeable enough to properly fix this issue myself. Does anyone know what to do to resolve this?

PerchunPak:
In the [future](https://github.com/NixOS/nixpkgs/issues/503391) version 26.05 (or already in nixos-unstable), NixOS [switched boot stage 1 to systemd](https://github.com/NixOS/nixpkgs/pull/435781), which means that the old setup now has to be rewritten to a systemd unit.

Resolves https://github.com/nix-community/impermanence/issues/320


README.org
```diff
@@ -156,6 +156,94 @@
...
+                  IFS=$'\n'
+                  for i in $(btrfs subvolume list -o "$1" | cut -f 9- -d ' '); do
+                      delete_subvolume_recursively "/btrfs_tmp/$i"
+                  done
+                  btrfs subvolume delete "$1"
+              }
+
+              for i in $(find /btrfs_tmp/old_roots/ -maxdepth 1 -mtime +30); do
+                  delete_subvolume_recursively "$i"
+              done
```
delliottxyz, line 216: BTRFS supports recursive delete using a mount option for some time now. Its a little cleaner than this, IMO.

I have a wipe script here that uses it: https://codeberg.org/delliott/nixos-config/src/branch/main/modules/nixos/common-btrfs-wipe.nix#L24-L55

The important bit is mount the subvol with the `user_subvol_rm_allowed` option, and then use the `-R` flag in the delete call.

Might be useful?

emilazy:
https://github.com/nix-community/impermanence/pull/321

cc @ElvishJerricco

README.org
PerchunPak: Yeah, it is very useful. I managed to simplify this script to

```sh
set -e
MOUNTDIR=/mnt
BTRFS_VOL=/dev/mapper/root_vg-root

echo "Mounting btrfs..."
mkdir -p $MOUNTDIR
mount -t btrfs -o subvol=/,user_subvol_rm_allowed $BTRFS_VOL $MOUNTDIR

if [[ -e $MOUNTDIR/root ]]; then
  echo "Moving existing root to the old_roots directory..."
  mkdir -p $MOUNTDIR/old_roots
  timestamp=$(date --date="@$(stat -c %Y $MOUNTDIR/root)" "+%Y-%m-%-d_%H:%M:%S")
  mv $MOUNTDIR/root "$MOUNTDIR/old_roots/$timestamp"
fi

echo "Deleting old subvolumes..."
for old_subvolume in $(find $MOUNTDIR/old_roots/ -maxdepth 1 -mtime +30); do
  echo "Deleting $old_subvolume"
  btrfs subvolume delete -R "$old_subvolume"
done

btrfs subvolume create $MOUNTDIR/root
umount $MOUNTDIR
echo "Done!"
```

I won't add it here, as it is out of scope, but feel free to make a new PR

lxp: I would additionally first delete old subvolumes and only then move the existing root to old_roots to avoid directly deleting the existing root if it was modified over 30 days ago.

README.org
```diff
@@ -156,6 +156,94 @@
...
+          options = [ "subvol=root" ];
+        };
+
+        boot.initrd.systemd = {
+          initrdBin = with pkgs; [
+            btrfs-progs
+            coreutils
+            util-linux
+            findutils
+          ];
```
ElvishJerricco, line 181: These are all already there by default.

README.org
```diff
@@ -156,6 +156,94 @@
...
+
+          services.prune-subvolumes = {
+            requiredBy = [ "initrd.target" ];
+            before = [ "local-fs-pre.target" ];
```
ElvishJerricco, line 185: I've advised `before = [ "local-fs-pre.target" ];` along with `after = [ "systemd-hibernate-resume.service" ];` in the past, but these days I tend to recommend `after = [ "local-fs-pre.target" ];` along with `before = [ "sysroot.mount" ];`. The reason is that `local-fs-pre.target` is about things that must happen before any file system is mounted, and that includes the mounting that gets done in this service. Yes, the most practical such thing is `systemd-hibernate-resume.service`, so ordering after that is good enough, but I think in principle it's better to stick to the rule about ordering after `local-fs-pre.target`, and then ordering before the things that we know we specifically want to block like the root fs.

ElvishJerricco (commented):
Hi, yea I posted a PR for using systemd stage 1 already (#321). I've commented the main technical differences, but I'd also note that I think it's simpler to just modify the existing section and tell users to enable systemd stage 1 on 25.11 rather than having two different versions in the README. The systemd stage 1 version should pretty much work for quite a few NixOS releases back by now, and encouraging people to migrate to it is a good idea considering it's the default now and scripted stage 1 will be removed in 26.11; these instructions are for new installs, so I don't think the potential compatibility issues with older configs is a problem.

PerchunPak:
I will close this in favour of https://github.com/nix-community/impermanence/pull/321, as the author (@ElvishJerricco) is a member of the [`@NixOS/systemd`](https://github.com/orgs/NixOS/teams/systemd) team and they surely know better than I am

